PR #10, opened 2019-04-12
Generalize address form conforms

dmnkdmnt:


cart/interfaces/controller/forms/basics.go
```diff
@@ -9,26 +9,26 @@ type (
...
-		Email            string `form:"email" validate:"required,email"`
+		City             string `form:"city" conform:"ucfirst,trim"`
+		State            string `form:"state" conform:"ucfirst,trim"`
+		RegionCode       string `form:"regionCode" conform:"trim"`
+		Country          string `form:"country" conform:"ucfirst,trim"`
+		CountryCode      string `form:"countryCode" conform:"trim"`
+		PhoneAreaCode    string `form:"phoneAreaCode" conform:"trim"`
+		PhoneCountryCode string `form:"phoneCountryCode" conform:"trim"`
+		PhoneNumber      string `form:"phoneNumber" conform:"trim"`
+		Email            string `form:"email" validate:"required,email" conform:"email"`
```
thebod: Is conform able to handle all email values? Including chinese, short ones, local ones, etc?

dmnkdmnt: it changes only to lower case and trims

PR #17, opened 2019-04-18
WIP: cart: Validate payment selection

carstendietrich:
Changes to the cart can lead to invalid PaymentSelections, therefore we need to validate the PaymentSelection after each operation. If the PaymentSelection is invalid (cart.GrandTotal != charges total) we should reset the PaymentSelection.

Furthermore we should check during UpdatePaymentSelection() if the supplied payment selection is set and valid.


cart/infrastructure/InMemoryBehaviour.go
```diff
@@ -327,7 +348,29 @@ func (cob *InMemoryBehaviour) ApplyVoucher(ctx context.Context, cart *domaincart
...
+
+func (cob *InMemoryBehaviour) isCurrentPaymentSelectionValid(ctx context.Context, cart *domaincart.Cart) bool {
+	return cob.isPaymentSelectionValid(ctx, cart, cart.PaymentSelection)
+}
+
+// isPaymentSelectionValid checks if the grand total of the cart matches the total of the supplied payment selection
+func (cob *InMemoryBehaviour) isPaymentSelectionValid(ctx context.Context, cart *domaincart.Cart, paymentSelection *domaincart.PaymentSelection) bool {
+	var chargePrices []domainPrice.Price
+	for _, charge := range paymentSelection.GetCharges().GetAllCharges() {
+		chargePrices = append(chargePrices, charge.Price)
```
danielpoe: Use charge Value..
Price can be in different currencies

PR #19, opened 2019-04-23
cart: Add event that fires when the payment selection was reset

carstendietrich:
If the payment selection is reset, we should dispatch an event to inform the frontend to generate a new payment selection.

fixed some linting errors in cartReceiver_test.go as well.

### DIFF
--- a/cart/application/eventPublishing.go
+++ b/cart/application/eventPublishing.go
@@ -33,11 +33,16 @@ type (
 		QtyAfter               int
 	}
 
+	// PaymentSelectionHasBeenResetEvent defines event properties
+	PaymentSelectionHasBeenResetEvent struct {
+	}
+
 	//EventPublisher - technology free interface to publish events that might be interesting for outside (Publish)
 	EventPublisher interface {
 		PublishAddToCartEvent(ctx context.Context, marketPlaceCode string, variantMarketPlaceCode string, qty int)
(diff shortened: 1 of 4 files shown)


cart/application/eventPublishing.go
```diff
@@ -33,11 +33,16 @@ type (
...
+	// PaymentSelectionHasBeenResetEvent defines event properties
+	PaymentSelectionHasBeenResetEvent struct {
+	}
+
 	//EventPublisher - technology free interface to publish events that might be interesting for outside (Publish)
 	EventPublisher interface {
 		PublishAddToCartEvent(ctx context.Context, marketPlaceCode string, variantMarketPlaceCode string, qty int)
 		PublishChangedQtyInCartEvent(ctx context.Context, item *cartDomain.Item, qtyBefore int, qtyAfter int, cartID string)
 		PublishOrderPlacedEvent(ctx context.Context, cart *cartDomain.Cart, placedOrderInfos cartDomain.PlacedOrderInfos)
+		PublishPaymentSelectionHasBeenResetEvent(ctx context.Context)
```
danielpoe: I think passing the cart reference makes sense and gives the event the correct context?

carstendietrich: I'll added the cart reference, please have a look

PR #26, opened 2019-05-13
cart: Add missing config struct to fix cart tests

carstendietrich:


cart/application/cartReceiver_test.go
```diff
@@ -549,7 +552,15 @@ func TestCartReceiverService_DecorateCart(t *testing.T) {
 
 func TestCartReceiverService_GetDecoratedCart(t *testing.T) {
 	authmanager := &authApplication.AuthManager{}
-	authmanager.Inject(flamingo.NullLogger{},nil,nil,nil)
+	authmanager.Inject(flamingo.NullLogger{}, nil, nil, &struct {
```
tessig: I would prefer the `Inject` method to be graceful if config is `nil` instead of having to pass an empty config struct if none is required.

PR #31, opened 2019-05-20
Improve datalayer

dylan-provost:


w3cdatalayer/application/factory.go
```diff
@@ -302,6 +311,10 @@ func (s Factory) getProductCategory(product productDomain.BasicProduct) *domain.
 func (s Factory) getProductInfo(product productDomain.BasicProduct) domain.ProductInfo {
 	baseData := product.BaseData()
 	retailerCode := baseData.RetailerCode
+
+	productName := baseData.Title
+	productName = s.regex.ReplaceAllString(baseData.Title, "-")
```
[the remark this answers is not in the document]
dylan-provost: It used to be private and was then created in the inject on suggestion from Thomas Neumann for potential performance issues...

w3cdatalayer/application/factory.go
```diff
@@ -65,6 +67,13 @@ func (s *Factory) Inject(
 	s.locale = config.Locale
 	s.defaultCurrency = config.DefaultCurrency
 	s.hashUserValues = config.HashUserValues
+
+	regexString := "[,|;|\\|]"
```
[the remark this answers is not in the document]
dylan-provost: This applies only to certain rules for Adobe data catcher on product names. We could escape more but it was not a requirement in the story (https://targetprocess.aoe.com/entity/194598-26-change-datalayer-data). Would you suggest to extend i?

w3cdatalayer/application/factory.go
```diff
@@ -65,6 +67,13 @@ func (s *Factory) Inject(
...
+	r, err := regexp.Compile(regexString)
+	if err != nil {
+		panic(err)
```
chetan-thapliyal: Do we really need a panic here? It'll ultimately result in crashing the whole app. Is this really the desired behavior?

w3cdatalayer/application/factory.go
```diff
@@ -302,6 +311,10 @@ func (s Factory) getProductCategory(product productDomain.BasicProduct) *domain.
 func (s Factory) getProductInfo(product productDomain.BasicProduct) domain.ProductInfo {
 	baseData := product.BaseData()
 	retailerCode := baseData.RetailerCode
+
+	productName := baseData.Title
```
chetan-thapliyal: This statement is redundant. productName is immediately overwritten in next statement.

w3cdatalayer/application/factory.go
dylan-provost: If that can answer your question, see the discussion with Thomas here https://gitlab.aoe.com/shared/i-love-flamingo/flamingo-commerce/merge_requests/58#note_44273

masmrlar: @chetan-thapliyal I agree with you that `panic` should always be avoided. 
In this case we got a pretty good reason for the panic: If we have an erratic regex string, the regex won't compile and we won't get the expected functionality, but lots of possible error logs. 

So a better way would be to just panic at this point - as an erratic regex is equivalent to broken code.

w3cdatalayer/application/factory.go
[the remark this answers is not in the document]
dylan-provost: @danielpoe is this still an open question?

PR #38, opened 2019-05-28
cart: Add delivery code to restrictors

carstendietrich:
Allow restrictors to restrict quantities based on delivery code.
This PR also adds Attributes to OrderItems to be more flexible in the future.


jeinfeldt:
Nice one, thanks for contributing

PR #40, opened 2019-05-31
In DEVMode: restore the last stored transaction flash data

begner:
In DEVMode: restore the last stored transaction flash data (orderdata) so a frontend developer can reload the successpage (without creating orders again and again) while working on success template

### DIFF
--- a/checkout/interfaces/controller/checkoutcontroller.go
+++ b/checkout/interfaces/controller/checkoutcontroller.go
@@ -84,6 +84,8 @@ type (
 		redirectToCartOnInvalideCart    bool
 		privacyPolicyRequired           bool
 
+		DevMode bool `inject:"config:debug.mode"`
+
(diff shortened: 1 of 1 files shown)


checkout/interfaces/controller/checkoutcontroller.go
```diff
@@ -84,6 +84,8 @@ type (
 		redirectToCartOnInvalideCart    bool
 		privacyPolicyRequired           bool
 
+		DevMode bool `inject:"config:debug.mode"`
```
tessig: Direct injections are deprecated. Please use `CheckoutController`'s `Inject` function and pass an anonymous config struct (like it is done in the `CartViewController` for example)

yesman82: Updated code to use the chainable injection method

PR #43, opened 2019-06-07
cart: Add option to remove applied voucher from cart

carstendietrich:
This PR allows the user to remove already applied vouchers from card.

### DIFF
--- a/cart/infrastructure/InMemoryBehaviour.go
+++ b/cart/infrastructure/InMemoryBehaviour.go
@@ -2,11 +2,12 @@ package infrastructure
 
 import (
 	"context"
-	"flamingo.me/flamingo-commerce/v3/cart/domain/events"
 	"fmt"
 	"math/big"
 	"math/rand"
 	"strconv"
+
+	"flamingo.me/flamingo-commerce/v3/cart/domain/events"
 
 	domaincart "flamingo.me/flamingo-commerce/v3/cart/domain/cart"
 	"flamingo.me/flamingo-commerce/v3/product/domain"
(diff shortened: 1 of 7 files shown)


cart/infrastructure/InMemoryBehaviour.go
```diff
@@ -354,6 +355,24 @@ func (cob *InMemoryBehaviour) ApplyVoucher(ctx context.Context, cart *domaincart
...
 }
 
+// RemoveVoucher removes a voucher from the cart
+func (cob *InMemoryBehaviour) RemoveVoucher(ctx context.Context, cart *domaincart.Cart, couponCode string) (*domaincart.Cart, domaincart.DeferEvents, error) {
+	for i, coupon := range cart.AppliedCouponCodes {
+		if coupon.Code == couponCode {
+			cart.AppliedCouponCodes[i] = cart.AppliedCouponCodes[len(cart.AppliedCouponCodes)-1]
+			cart.AppliedCouponCodes[len(cart.AppliedCouponCodes)-1] = domaincart.CouponCode{}
+			cart.AppliedCouponCodes = cart.AppliedCouponCodes[:len(cart.AppliedCouponCodes)-1]
+		}
```
jeinfeldt: We assume that only one coupon code matches, right? In that case you could insert a break after the removal of the element, as there is no need to iterate through the codes further

cart/module.go
```diff
@@ -158,6 +158,10 @@ func (r *routes) apiRoutes(registry *web.RouterRegistry) {
 	registry.HandlePost("cart.api.applyVoucher", r.apiController.ApplyVoucherAndGetAction)
 	registry.HandlePut("cart.api.applyVoucher", r.apiController.ApplyVoucherAndGetAction)
 
+	registry.Route("/api/cart/removevoucher", `cart.api.removeVoucher(couponCode)`)
+	registry.HandlePost("cart.api.removevoucher", r.apiController.RemoveVoucherAndGetAction)
+	registry.HandleDelete("cart.api.removevoucher", r.apiController.RemoveVoucherAndGetAction)
+
```
jeinfeldt: cart.api.removeVoucher is declared in camel case but used in lower case

cart.ape.removevoucher. If an exact match is necessary, please adjust

jeinfeldt:
Thanks for contributing!

cart/module.go
carstendietrich: I updated the handle name accordingly.

cart/infrastructure/InMemoryBehaviour.go
carstendietrich: Good point, I added the break.

jeinfeldt:
Thanks for contributing! 👍

PR #49, opened 2019-06-21
Enhance cart behaviour contract with giftcard functionality

jeinfeldt:
Hola! I would like to add giftcard functionality to the cart contract. They should be handled separately from vouchers, as their implementation could be different. My suggestion is to enhance the cart behaviour interface with functions analogue to vouchers.


cart/domain/cart/cartServicePorts.go
```diff
@@ -47,6 +47,8 @@ type (
 		UpdateDeliveryInfoAdditionalData(ctx context.Context, cart *Cart, deliveryCode string, additionalData *AdditionalData) (*Cart, DeferEvents, error)
 		ApplyVoucher(ctx context.Context, cart *Cart, couponCode string) (*Cart, DeferEvents, error)
 		RemoveVoucher(ctx context.Context, cart *Cart, couponCode string) (*Cart, DeferEvents, error)
+		ApplyGiftCard(ctx context.Context, cart *Cart, giftCardCode string) (*Cart, DeferEvents, error)
+		RemoveGiftCard(ctx context.Context, cart *Cart, giftCardCode string) (*Cart, DeferEvents, error)
```
danielpoe: Next time we can think of adding an additional iface (GiftCardBehaviour) - and in the applications service check if the cartbehaviour also implements this iface..

Makes it less troublesome for other secondary ports to implement methods

PR #50, opened 2019-06-24
cart: Combined Voucher and Gift Card Code API Endpoint

jadhub:


cart/interfaces/controller/cartapicontroller.go
```diff
@@ -145,6 +145,53 @@ func (cc *CartAPIController) ApplyGiftCardAndGetAction(ctx context.Context, r *w
...
+	if err != nil {
+		result.SetError(err, "combined_cart_error")
+		response := cc.responder.Data(result)
+		response.Status(500)
+
+		return response
+	}
+
+	if len(currentCart.AppliedCouponCodes) == 0 {
+		// no coupon at cart, try to apply voucher
```
jeinfeldt: I think we would like to overwrite an existing voucher code, so we leave out this if.
Apply voucher code -> success -> return
Apply voucher code -> failure -> Apply giftcard code

cart/interfaces/controller/cartapicontroller.go
```diff
@@ -145,6 +145,30 @@ func (cc *CartAPIController) ApplyGiftCardAndGetAction(ctx context.Context, r *w
 	return cc.handlePromotionAction(ctx, r, "giftcard_error", cc.cartService.ApplyGiftCard)
 }
 
+// ApplyCombinedVoucherGift applies a given first to the voucher API Endpoint (if no voucher code is already set)
```
carstendietrich: since we don't know how the underlying cart adapter works we should adjust this comment.

jadhub: moved to cartservice interface

PR #54, opened 2019-06-27
cart: add json de-/serialization of PaymentSplit

jeinfeldt:
Currently the paymentsplit cannot be parsed to json as the key in the map is a complex object. I suggest to parse the information in this object to a string key to provide marshalling and unmarshalling behaviour.

In order to migitate overwrite of keys due to empty methods or chargetypes, I suggest to throw an error in case such a value is encountered.

### DIFF
--- a/cart/domain/cart/paymentselection.go
+++ b/cart/domain/cart/paymentselection.go
@@ -1,7 +1,15 @@
 package cart
 
 import (
+	"encoding/json"
+	"errors"
+	"strings"
+
 	price "flamingo.me/flamingo-commerce/v3/price/domain"
+)
+
+const (
+	splitQualifierSeperator = "___"
 )
 
 type (
@@ -20,6 +28,9 @@ type (
 		ChargeType string
 		Method     string
 	}
+
+	// PaymentSplitJSON - representation of complete payment split
+	PaymentSplitJSON map[string]price.Charge
 
 	//PaymentSplit - the Charges qualified by Type and PaymentMethod
 	PaymentSplit map[SplitQualifier]price.Charge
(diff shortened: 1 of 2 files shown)


cart/domain/cart/paymentselection.go
```diff
@@ -21,6 +29,9 @@ type (
 		Method     string
 	}
 
+	// PaymentSplitJSON - representation of complete payment split
+	PaymentSplitJSON map[string]price.Charge
```
carstendietrich: Do we actually need this as a public struct? Looks more like a helper struct that is only used in the marshalling functions..

cart/domain/cart/paymentselection_test.go
```diff
@@ -41,6 +42,106 @@ func TestPrice_MarshalBinaryForGob(t *testing.T) {
...
+			},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			split := cart.PaymentSplit{}
+			if err := split.UnmarshalJSON(tt.args.data); (err != nil) != tt.wantErr {
+				t.Errorf("PaymentSplit.UnmarshalJSON() error = %v, wantErr %v", err, tt.wantErr)
+			}
```
carstendietrich: I would add an additional check to verify that the content of the unmarshalled split is actually correct.

jeinfeldt: Good point, enahnced test

cart/domain/cart/paymentselection.go
jeinfeldt: Removed helper struct, thanks for the hint

cart/domain/cart/paymentselection.go
```diff
@@ -1,9 +1,17 @@
...
 import (
+	"encoding/json"
+	"errors"
+	"strings"
+
 	price "flamingo.me/flamingo-commerce/v3/price/domain"
 )
 
+const (
+	splitQualifierSeperator = "___"
```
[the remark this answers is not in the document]
jeinfeldt: Fixed

PR #58, opened 2019-07-03
cart: always reserve an order id if requested

masmrlar:
The CartService must always reserve a (new) orderID fetched from the PlaceOrderService if requested.

The reciever ReserveOrderIDAndSave must not check if an order ID is already set on the cart.

### DIFF
--- a/cart/application/cartService.go
+++ b/cart/application/cartService.go
@@ -729,6 +729,7 @@ func (cs *CartService) DeleteCartInCache(ctx context.Context, session *web.Sessi
 
 // ReserveOrderIDAndSave reserves order id by using the PlaceOrder behaviour, sets and saves it on the cart.
 // Each call of this method reserves a new order ID, even if it is already set on the cart.
+// You may want to use this before proceeding with payment to ensure having a useful reference in the payment processing
 func (cs *CartService) ReserveOrderIDAndSave(ctx context.Context, session *web.Session) (*cartDomain.Cart, error) {
 	if cs.placeOrderService == nil {
 		return nil, errors.New("No placeOrderService registered")
--- a/cart/application/cartService.go
+++ b/cart/application/cartService.go
@@ -728,8 +728,7 @@ func (cs *CartService) DeleteCartInCache(ctx context.Context, session *web.Sessi
 }
 
 // ReserveOrderIDAndSave reserves order id by using the PlaceOrder behaviour, sets and saves it on the cart.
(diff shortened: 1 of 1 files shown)


cart/application/cartService.go
```diff
@@ -728,8 +728,7 @@ func (cs *CartService) DeleteCartInCache(ctx context.Context, session *web.Sessi
 }
 
 // ReserveOrderIDAndSave reserves order id by using the PlaceOrder behaviour, sets and saves it on the cart.
-// If the cart already holds a reserved order id no set/save is performed and the existing cart is returned.
-// You may want to use this before proceeding with payment to ensure having a useful reference in the payment processing
```
tessig: This last hint should stay.

masmrlar: You're right - we should not remove that comment.

danielpoe:
> The CartService must always reserve a (new) orderID fetched from the PlaceOrderService if requested.
> 
> The reciever ReserveOrderIDAndSave must not check if an order ID is already set on the cart.

@masmrlar can you explain why?

masmrlar:
> > The CartService must always reserve a (new) orderID fetched from the PlaceOrderService if requested.
> > The reciever ReserveOrderIDAndSave must not check if an order ID is already set on the cart.
> 
> @masmrlar can you explain why?

Method is `ReserverOrderIDAndSave` leading to the expectation that a new orderID is reserved. 
The current implementation drops the reservation if one is already set, which may lead to erratic behaviour in checkouts expecting to call this method to reserve a fresh order ID.

danielpoe:
> > > The CartService must always reserve a (new) orderID fetched from the PlaceOrderService if requested.
> > > The reciever ReserveOrderIDAndSave must not check if an order ID is already set on the cart.
> > 
> > 
> > @masmrlar can you explain why?
> 
> Method is `ReserverOrderIDAndSave` leading to the expectation that a new orderID is reserved.
> The current implementation drops the reservation if one is already set, which may lead to erratic behaviour in checkouts expecting to call this method to reserve a fresh order ID.

As discussed:
The intent is that calls to this method are idempotent.
Once a cart has a orderid reserved it ist the future orderid for this cart - until the cart is deleted or placed. This is because the reserved order id might be used by external systems before placing the order and therefore it should stay as a second reference.

We can add a "Forced" method if one want to avoid the functionality for some reason.

jeinfeldt:
PR is outdated, will therefore be closed

PR #60, opened 2019-07-08
cart: Resctructure discounts

jeinfeldt:
Currently Discounts are available on Item level (ItemDiscounts) but not on a delivery or a cart level. These information may be useful for the client, therefore I suggest a new generic structure of Discounts that are available on Item, Delivery and Cart level. ~~The discounts are merged by their title and a newly introduced type~~. 

The discounts are merged by their campaign code, which is defined as a code that is a underlying identifier.

### DIFF
--- a/cart/domain/cart/discount.go
+++ b/cart/domain/cart/discount.go
@@ -145,6 +145,7 @@ func (discounts AppliedDiscounts) ByCampaignCode(campaignCode string) AppliedDis
 		return discount.CampaignCode == campaignCode
 	}
 	return discounts.filter(f)
+
 }
 
 // ByType filter AppliedDiscounts based on type
--- a/cart/domain/cart/discount.go
+++ b/cart/domain/cart/discount.go
@@ -16,24 +16,27 @@ type (
 	// WithDiscount interface for a cart that is able to handle discounts
 	WithDiscount interface {
 		HasAppliedDiscounts() bool
-		AggregateDiscounts() []*AppliedDiscount
+		CollectDiscounts() []*AppliedDiscount
 	}
+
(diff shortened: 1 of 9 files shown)


cart/domain/cart/discount.go
```diff
@@ -0,0 +1,165 @@
...
+)
+
+type (
+	// AppliedDiscount value object - generic reference for a discount
+	AppliedDiscount struct {
+		Code    string       // unique code of discount
+		Title   string       // readable name of discount
+		Applied domain.Price // how much of the discount has been subtracted from cart price
+		Type    string       // to distinguish between discounts
+	}
```
jeinfeldt: Code - fachlicher schlüssel der Kampagne (CampaignCode)
Type - ggf zusammen mit Code der Schlüssel 
--  Const Type?  Coupon 
CouponCode - wenn Type Coupon
Applied - Document that negative price should be used
Title -> Rename to Label

IsItemRelated bool


=> Dann ItemDiscount Value object löschen

jeinfeldt: Remove error 
-> In case add returns error, break and log

cart/domain/decorator/cartDecorator.go
```diff
@@ -54,6 +56,20 @@ func (df *DecoratedCartFactory) Inject(
 	df.logger = logger
 }
 
+// Inject dependencies
+func (dci *DecoratedCartItem) Inject(
```
tessig: Do we ever get Cart Items through DI?

jeinfeldt: Good point! I removed the dependency injection and added the logger via factory methods

PR #62, opened 2019-07-12
Bugfix: Allow filters with multiple values

yesman82:
For example, selecting multiple delivery methods or brands was not possible.

With this fix it's allowed to either
- send an filter list with strings as values, like { 'brand': 'apple' } (the old way for backwards compatibility
- or send a list with an array of strings, like { 'brand': ['apple', 'samsung'] }

### DIFF
--- a/product/interfaces/templatefunctions/findProducts.go
+++ b/product/interfaces/templatefunctions/findProducts.go
@@ -37,16 +37,14 @@ func (tf *FindProducts) Func(ctx context.Context) interface{} {
 		config - map with certain keys - used to specify the searchRequest better
 	*/
 	return func(namespace string, configs ...pugjs.Map) *application.SearchResult {
-		var searchConfig, filterConstrains map[string]string
-		var keyValueFilters map[string][]string
+		var searchConfig = make(map[string]string)
+		var filterConstrains = make(map[string]string)
+		var keyValueFilters = make(map[string][]string)
 
 		if len(configs) > 0 {
 			searchConfig = configs[0].AsStringMap()
-		} else {
-			searchConfig = make(map[string]string)
 		}
 
-		keyValueFilters = make(map[string][]string)
 		if len(configs) > 1 {
 			for key, value := range configs[1].AsStringIfaceMap() {
 				if str, ok := value.(pugjs.String); ok { // allow string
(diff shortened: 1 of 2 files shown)


product/interfaces/templatefunctions/findProducts.go
```diff
@@ -33,25 +34,31 @@ func (tf *FindProducts) Func(ctx context.Context) interface{} {
...
+	return func(namespace string, configs ...pugjs.Map) *application.SearchResult {
+		var searchConfig, filterConstrains map[string]string
+		var keyValueFilters map[string][]string
 
 		if len(configs) > 0 {
-			searchConfig = configs[0]
+			searchConfig = configs[0].AsStringMap()
 		} else {
 			searchConfig = make(map[string]string)
 		}
```
jeinfeldt: Suggestion: Initialise searchConfig as a map instead of just declaring it and remove the else case
I think this provides better readability as there is just a default case and an if

product/interfaces/templatefunctions/findProducts.go
```diff
@@ -33,25 +34,31 @@ func (tf *FindProducts) Func(ctx context.Context) interface{} {
...
+				}
+			}
 		}
 
 		if len(configs) > 2 {
-			filterConstrains = configs[2]
+			filterConstrains = configs[2].AsStringMap()
 		} else {
 			filterConstrains = make(map[string]string)
 		}
```
jeinfeldt: Same as first comment, I would suggest to remove the else case and init filterConstraints at the start of the function

product/interfaces/templatefunctions/findProducts.go
```diff
@@ -33,27 +34,29 @@ func (tf *FindProducts) Func(ctx context.Context) interface{} {
...
 		widgetName - used to namespace widget - in case we need pagination
-		config - map with certain keys - used to specifiy th searchRequest better
+		config - map with certain keys - used to specify the searchRequest better
 	*/
-	return func(namespace string, configs ...map[string]string) *application.SearchResult {
-		var searchConfig, keyValueFilters, filterConstrains map[string]string
+	return func(namespace string, configs ...pugjs.Map) *application.SearchResult {
+		var searchConfig = make(map[string]string)
+		var filterConstrains = make(map[string]string)
+		var keyValueFilters = make(map[string][]string)
```
jeinfeldt: To init you can simply put:
searchConfig := make(map[string]string)
etc.
You do not need the var

PR #63, opened 2019-07-16
price: add handling of negative values in split in payables

jeinfeldt:
Currently split in payables does not function correctly when called with negative values as the loop 

`	
for i := 0; i < int(splittedAmountModulo); i++ {
    splittedAmounts[i] = splittedAmounts[i] + 1
}
`
Does not move around the overflowing cent.
Therefore i suggest to multiply negative values with -1 to achieve the correct behaviour.


price/domain/price_test.go
```diff
@@ -249,3 +249,15 @@ func TestCharges_Add(t *testing.T) {
 		Type:  "main",
 	}, charge)
 }
+
+func TestPrice_SplitInPayables_NegativeValue(t *testing.T) {
```
carstendietrich: We could add this edge case to the TestPrice_SplitInPayables so that all test data related to the function are in one test func.

PR #65, opened 2019-07-16
cart: add event that is dispatched before an order is placed

carstendietrich:
Currently we only have an event after the order is placed. Maybe I want to check / do stuff right before the order is placed.

### DIFF
--- a/cart/application/cartReceiver_test.go
+++ b/cart/application/cartReceiver_test.go
@@ -8,7 +8,6 @@ import (
 	"flamingo.me/flamingo-commerce/v3/cart/domain/placeorder"
 	"flamingo.me/flamingo-commerce/v3/cart/domain/validation"
 
-
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


danielpoe:
What would be a concrete use case?
Why not add a payload to the event?

carstendietrich:
Concrete use case would be adding additional attributes to the cart before placing it. There are hooks for that, but they only get triggered on certain cart actions which make it hard to ensure correct attributes if they depend on the cart content. The event doesn't really need additional payload since you can fetch the cart from the context and update it.

danielpoe:
I think an event should be self contained - telling a bit about the ressource where it appeared.
(fetching the cart via sie effect might in theory give you something else in a concurrent system)

But still expecting the event to trigger via side effect a change on the cart that will be placed some lines after the event is triggered feels strange.

Events are meant for listing and not for changing something on the ressource via sideeffect

Currently we have a Hook for adding attributes directly in the Adapter>Mapper of flamingo_om3 - Maybe we can either:
- use simelar hook in the Adapter where we need it..
- or add an explicit Hook in flamingo_commerce if we think that should be a common behaviour that "someone" should modify the cart just before it is placed..

carstendietrich:
Jeah thats sounds reasonable I guess my solution was kinda rushed. Im closing this PR for now and come up with a better solution.

PR #66, opened 2019-07-16
Add sort order field for discounts

thehl:


cart/domain/cart/discount.go
```diff
@@ -13,6 +14,7 @@ type (
 		Applied       domain.Price // how much of the discount has been subtracted from cart price, IMPORTANT: always negative
 		Type          string       // to distinguish between discounts
 		IsItemRelated bool         // flag indicating if the discount is applied due to item in cart
+		SortOrder     int          // sort order for discounts, from low to high
 	}
```
jeinfeldt: Could you adjust the docstring to something like:
"indicates in which order discount have been applied, low value has been applied before high value"

PR #72, opened 2019-07-26
Allow filters with multiple values

dylan-provost:
Refactor the FindProducts function

### DIFF
--- a/product/Readme.md
+++ b/product/Readme.md
@@ -102,7 +102,7 @@ findProducts is a template function that returns a search result to show product
         * `query`: Optional - the search string that a "human" might have entered to filter the search
         * `pageSize`, `page`: Optional - set the page and the pageSize (for pagination)
         * `sortBy`, `sortDirection` (`A`/`D`): Optional - set the field that should be used to sort the search result
-    * `keyValueFilters`: A map of key values that are used as additional keyValue Filters in the searchRequest
+    * `keyValueFilters`: Optional - A map of key values that are used as additional keyValue Filters in the searchRequest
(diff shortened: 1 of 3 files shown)


dylan-provost:
> Please also update the Readme with an example

Since there is nothing new it was just a fix and a refactoring, I just fixed the typos in the Readme

danielpoe:
> > Please also update the Readme with an example
> 
> Since there is nothing new it was just a fix and a refactoring, I just fixed the typos in the Readme

The title is saying "Allow filters with multiple values" - so I it sounds like an extended feature??

dylan-provost:
> > > Please also update the Readme with an example
> > 
> > 
> > Since there is nothing new it was just a fix and a refactoring, I just fixed the typos in the Readme
> 
> The title is saying "Allow filters with multiple values" - so I it sounds like an extended feature??

I think the first implementation just completely missed that point in the first place. URLs can obviously contain multiple filters, and we should be able to filter all of them, that is why the filters are documented as an map of strings. But I don't know what the Readme should say more than it says right now. Any idea?

danielpoe:
> > > > Please also update the Readme with an example
> > > 
> > > 
> > > Since there is nothing new it was just a fix and a refactoring, I just fixed the typos in the Readme
> > 
> > 
> > The title is saying "Allow filters with multiple values" - so I it sounds like an extended feature??
> 
> I think the first implementation just completely missed that point in the first place. URLs can obviously contain multiple filters, and we should be able to filter all of them, that is why the filters are documented as an map of strings. But I don't know what the Readme should say more than it says right now. Any idea?

Ok then I misuderstood the title. Thanks for clarification

PR #73, opened 2019-07-29
cart: remove zero charges from payment selection

jeinfeldt:
Currently for the calculation of splitted charges we need zero charges. These charges do not make sense for a clients payment (0 charges cannot be payed). 
Therefore before the payment selection is generated for the client I suggest to filter out 
zero charges.

### DIFF
--- a/cart/domain/cart/paymentselection.go
+++ b/cart/domain/cart/paymentselection.go
@@ -94,13 +94,13 @@ func NewDefaultPaymentSelection(gateway string, chargeTypeToPaymentMethod map[st
 	}
 	result, err := newPaymentSelectionWithGiftCard(gateway, chargeTypeToPaymentMethod, pricedItems, giftCards)
 	// filter out zero charges from here on out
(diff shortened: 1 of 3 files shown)


cart/domain/cart/paymentselection.go
```diff
@@ -57,13 +57,16 @@ type (
 	// PaymentSplitService enables the creation of a PaymentSplitByItem following different payment methods
 	PaymentSplitService struct{}
 
+	// builderAddFunc is a function used by builder to add items
+	// function which corresponds to builder adddX function (addCartItem, addShipping, addTotal)
```
thehl: Typo in comment. adddX vs addX.

jeinfeldt: Thanks for the hint, fixed

jeinfeldt:
> Looks good, i'll added some changes to make the function public so if you create your own payment selection you can use that helper. Please have a look if that matches your intention.

Cool.

PR #74, opened 2019-07-31
checkout: enable cart to be passed for order placement

jeinfeldt:
Currently as a client I can only place  a cart with payment processing which is taken from my cache or freshly fetched by my cartReceiverService implementation.

I suggest to provide a function to explicitly pass a cart for order placement and payment processing.


checkout/application/orderService.go
```diff
@@ -268,12 +300,14 @@ func (os *OrderService) CurrentCartPlaceOrderWithPaymentProcessing(ctx context.C
...
-func (os *OrderService) GetContactMail(cart cart.Cart) string {
-	//Get Email from either the cart
-	shippingEmail := cart.GetMainShippingEMail()
-	if shippingEmail == "" {
-		shippingEmail = cart.BillingAdress.Email
-	}
-	return shippingEmail
+// createFreshContext create a fresh context from a given context
+// that will not be canceled if original context is canceled by client
+func (os *OrderService) createFreshContext(ctx context.Context) context.Context {
```
carstendietrich: Can we rename this in something like generateNewBackgroundContext?

PR #79, opened 2019-08-09
payment/checkout: Enable payment / checkout to early place order  and show a robust payment page

carstendietrich:
Currently the order is placed after the payment process is finished. Due to cross device payment options some payment adapters could demand an early order placement and flamingo-commerce should offer an polling option ask the payment adapter in which state the payment flow is. According to that payment status certain messages can be displayed to the customer.

To avoid polluting the SubmitCheckoutAction/PlaceOrderAction this handling should happen in a separate action which should display an intermediate payment page. This page can then react to the current payment flow status and display certain messages through polling or redirecting to the SuccessAction if the payment adapter tells us to do so.

### DIFF
--- a/checkout/application/orderService.go
+++ b/checkout/application/orderService.go
@@ -337,6 +337,7 @@ func (os *OrderService) preparePlaceOrderInfo(ctx context.Context, currentCart c
 	placeOrderInfo := &PlaceOrderInfo{
 		ContactEmail: email,
 		PlacedOrders: placedOrderInfos,
+		Cart:         currentCart,
 	}
 
 	for _, transaction := range cartPayment.Transactions {
--- a/checkout/application/orderService.go
+++ b/checkout/application/orderService.go
@@ -38,7 +38,9 @@ type (
 		PaymentInfos []PlaceOrderPaymentInfo
 		PlacedOrders placeorder.PlacedOrderInfos
 		ContactEmail string
+		Cart         cart.Cart
 	}
+
 	//PlaceOrderPaymentInfo holding payment infos
 	PlaceOrderPaymentInfo struct {
 		Gateway         string
@@ -51,8 +53,11 @@ type (
 )
 
 const (
-	//PaymentFlowStandardCorrelationID - used as correlationid for the start of the payment (session scoped)
+	// PaymentFlowStandardCorrelationID used as correlationid for the start of the payment (session scoped)
 	PaymentFlowStandardCorrelationID = "checkout"
+
+	// LastPlacedOrderSessionKey is the session key for storing the last placed order
+	LastPlacedOrderSessionKey = "orderservice_last_placed"
 )
 
 var orderFailedStat = stats.Int64("flamingo-commerce/orderfailed", "my stat records 1 occurences per error", stats.UnitDimensionless)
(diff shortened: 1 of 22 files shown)


payment/domain/payment.go
```diff
@@ -11,10 +9,25 @@ type (
...
 
-	// FlowResult contains an url and a type to use to start a flow
+	// FlowResult contains information about a newly started flow
 	FlowResult struct {
-		URL  *url.URL
-		Type string
+		// Status of the payment flow
+		Status string
+		// PlaceOrder indicates if the order should be placed with the beginning of the payment flow
+		PlaceOrder bool
```
zohmi: I would rename PlaceOrder. The name sounds as if it would only be placed if this one is set :)

carstendietrich: Good point, I renamed to prop to EarlyPlaceOrder to be more speaking.

payment/interfaces/offlinePaymentGateway.go
```diff
@@ -77,21 +77,37 @@ func (o *OfflineWebCartPaymentGateway) StartFlow(ctx context.Context, currentCar
 	return &domain.FlowResult{}, nil
 }
 
-//StartWebFlow for offline payment requires not much - just redirect to the returnUrl :-)
-func (o *OfflineWebCartPaymentGateway) StartWebFlow(ctx context.Context, currentCart *cartDomain.Cart, correlationID string, returnURL *url.URL) (web.Result, error) {
+// FlowStatus for offline payment is always finished
+func (o *OfflineWebCartPaymentGateway) FlowStatus(ctx context.Context, cart *cartDomain.Cart, correlationID string) (*domain.FlowStatus, error) {
+	return &domain.FlowStatus{
+		Status: "complete",
+		Action: "SHOW_SUCCESS_PAGE",
```
jeinfeldt: I think this should return one of the newly introduced PaymentStatus constants, otherwise PaymentAction will handle default case. Unless this is intended

carstendietrich: fixed, thanks!

checkout/application/orderService.go
```diff
@@ -295,16 +349,6 @@ func (os *OrderService) placeOrderWithPaymentProcessing(ctx context.Context, dec
 			CreditCardInfo:  transaction.CreditCardInfo,
 		})
 	}
```
jeinfeldt: If i see correctly the cart is only used to fetch the e-mail? Should we not add the cart to the PlacedOrderInfo?

checkout/interfaces/controller/checkoutcontroller.go
```diff
@@ -244,26 +245,30 @@ func (cc *CheckoutController) PlaceOrderAction(ctx context.Context, r *web.Reque
...
-
-	if err != nil {
-		cc.logger.WithContext(ctx).WithField("subcategory", "checkoutError").WithField("errorMsg", err.Error()).Error(fmt.Sprintf("place order failed: cart id: %v / total-amount: %v", decoratedCart.Cart.EntityID, decoratedCart.Cart.GrandTotal()))
-		if paymentError, ok := err.(*paymentDomain.Error); ok {
-			if paymentError.ErrorCode == paymentDomain.ErrorPaymentAbortedByCustomer {
-				return cc.showCheckoutFormAndHandleSubmit(ctx, r, "checkout/checkout")
-			}
-			if cc.showReviewStepAfterPaymentError && !cc.skipReviewAction {
-				return cc.showReviewFormWithErrors(ctx, *decoratedCart, paymentError)
+		if err != nil {
```
jeinfeldt: I think you can just have the assignment of the placeOrderInfo in the if else 
and move the err evaluation up.
If we also assign the err for lastPlaceOrderInfo it is always evaluated and logged in case of an error, regardless what kind of error

The fetch of the placeOrderInfo could then also be moved to a private function

checkout/interfaces/controller/checkoutcontroller.go
```diff
@@ -625,3 +637,36 @@ func (cc *CheckoutController) checkTermsAndPrivacyPolicy(r *web.Request) (bool,
...
+	if err != nil {
+		cc.logger.Warn("couldn't get last placed order from orderService:", err)
+		return nil, err
+	}
+
+	return lastPlacedOrder, nil
+}
+
+// lastPlacedOrCurrentCart returns the decorated cart of the last placed order if there is one if not return the current cart
+func (cc *CheckoutController) lastPlacedOrCurrentCart(ctx context.Context) (*decorator.DecoratedCart, error) {
```
jeinfeldt: Nice!

payment/interfaces/controller/paymentapicontroller.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		pc.logger.Warn("couldn't get last placed order from orderService:", err)
+		return nil, err
+	}
+
+	return lastPlacedOrder, nil
+}
+
+// lastPlacedOrCurrentCart returns the decorated cart of the last placed order if there is one if not return the current cart
+func (pc *PaymentAPIController) lastPlacedOrCurrentCart(ctx context.Context) (*decorator.DecoratedCart, error) {
```
carstendietrich: Make public and move to order service to avoid duplicate code in checkoutcontroller / paymentapicontroller.

carstendietrich: done.

PR #80, opened 2019-08-15
204496: Fix promos not affecting shipping cost

ChristopherSchwab:
Replaced ShippingItem.DiscountAmount with ShippingItem.AppliedDiscounts.
ShippingItem not implements the WithDiscount interface.

### DIFF
--- a/cart/domain/cart/discount.go
+++ b/cart/domain/cart/discount.go
@@ -44,14 +44,9 @@ func (c *Cart) MergeDiscounts() (AppliedDiscounts, error) {
 	}
 	// collect different discounts on item level
 	var err error
-	var discounts AppliedDiscounts
 	collection := make(map[string]*AppliedDiscount)
 	for _, delivery := range c.Deliveries {
-		discounts, err = (&delivery).MergeDiscounts()
-		if err != nil {
-			return nil, err
-		}
-		collection, err = mapDiscounts(collection, discounts)
+		collection, err = mapDiscounts(collection, &delivery)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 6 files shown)


cart/domain/cart/discount.go
```diff
@@ -111,6 +116,25 @@ func (i *Item) HasAppliedDiscounts() (bool, error) {
...
+// MergeDiscounts parses discounts of a shipping shipping
+// All discounts with the same type and title are aggregated and returned as one with a summed price
+func (s *ShippingItem) MergeDiscounts() (AppliedDiscounts, error) {
+	sort.SliceStable(s.AppliedDiscounts, func(x, y int) bool {
+		return s.AppliedDiscounts[x].SortOrder < s.AppliedDiscounts[y].SortOrder
+	})
+
+	return s.AppliedDiscounts, nil
+}
+
```
jeinfeldt: You could try to use a more abstract private function for MergeDiscounts() for Items and Shipping Items
as they basically have the same code now

cart/domain/cart/discount.go
```diff
@@ -184,3 +208,16 @@ func (discounts AppliedDiscounts) filter(filterFunc func(AppliedDiscount) bool)
 	}
 	return result
 }
+
```
jeinfeldt: What about the idea the provide a force sum?

cart/domain/cart/discount.go
```diff
@@ -111,6 +116,25 @@ func (i *Item) HasAppliedDiscounts() (bool, error) {
...
+}
+
+// HasAppliedDiscounts checks whether there are any discounts currently applied to the cart affecting shipping cost
+func (s *ShippingItem) HasAppliedDiscounts() (bool, error) {
+	discounts, err := s.MergeDiscounts()
+	if err != nil {
+		return false, err
+	}
+	return len(discounts) > 0, nil
+}
```
jeinfeldt: Same as above but for HasAppliedDiscounts()

cart/domain/cart/discount.go
jeinfeldt: Discussed, not so useful

cart/domain/cart/discount_test.go
```diff
@@ -423,6 +423,109 @@ func TestItem_HasDiscounts(t *testing.T) {
 	}
 }
```
jeinfeldt: Please add a test with a complete cart including shipping and regular items to show that shipping items are now includes for the mergeDiscount on cart level

cart/domain/cart/discount.go
```diff
@@ -128,12 +164,7 @@ func mapToSlice(collection map[string]*AppliedDiscount) []AppliedDiscount {
...
 
 // mapDiscounts map type + title of discount to corresponding discount
-func mapDiscounts(result map[string]*AppliedDiscount, discountable WithDiscount) (map[string]*AppliedDiscount, error) {
-	discounts, err := discountable.MergeDiscounts()
-	// in case discounts cannot be collected, stop execution
-	if err != nil {
-		return nil, err
-	}
+func mapDiscounts(result map[string]*AppliedDiscount, discounts AppliedDiscounts) (map[string]*AppliedDiscount, error) {
 	for _, discount := range discounts {
```
jeinfeldt: I think this code can still be left in the function by passing an interface. Otherwise this code is the same for every place it is called

PR #87, opened 2019-08-26
cart: Add convenience function to calculate cart shipping cost

ChristopherSchwab:


cart/domain/cart/cart.go
```diff
@@ -327,6 +327,20 @@ func (c Cart) SumShippingNet() domain.Price {
...
+	for _, del := range c.Deliveries {
+		discount, _ := del.ShippingItem.AppliedDiscounts.Sum()
+		prices = append(prices, del.ShippingItem.PriceNet.ForceAdd(discount))
+	}
+
+	price, _ := domain.SumAll(prices...)
+
+	return price
+}
+
```
jeinfeldt: Could you maybe provide a test showing what price you expect given a cart with items and shipping items from this function?

ChristopherSchwab: Added a test as requested.

PR #94, opened 2019-09-02
checkout: fix module test

jeinfeldt:
Tests are failing due to missing config values

### DIFF
--- a/checkout/module_test.go
+++ b/checkout/module_test.go
@@ -1,14 +1,23 @@
 package checkout_test
 
 import (
-	"testing"
+"testing"
 
-	"flamingo.me/flamingo-commerce/v3/checkout"
-	"flamingo.me/dingo"
+"flamingo.me/flamingo-commerce/v3/checkout"
+"flamingo.me/dingo"
+"flamingo.me/flamingo/v3/framework/config"
 )
 
 func TestModule_Configure(t *testing.T) {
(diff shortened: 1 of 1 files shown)


checkout/module_test.go
```diff
@@ -1,14 +1,23 @@
...
 )
 
 func TestModule_Configure(t *testing.T) {
-	if err := dingo.TryModule(new(checkout.Module)); err != nil {
+	cfgModule := &config.Module{
+		Map: new(checkout.Module).DefaultConfig(),
+	}
+
+	cfgModule.Map["session.backend"] = ""
+	cfgModule.Map["oauth.useFake"] = true
```
danielpoe: only fixable that way?
Feels a bit strange that the checkout modul needs to know this internals

jeinfeldt: This is due to the fact that the checkout module depends on the cart and the cart module depends on the oauth module from core (which relies on these flags). I am unsure how to resolve these dependencies correctly in the module test without providing these flags

I think the only other option is to merge all config maps of the dependent modules together.

thebod:
Hi, I'm sorry, i accidentially pushed that fix to master.
This will be solved/cleaned up with the improved flamingo core config.

I'll close this, sorry!

checkout/module_test.go
thebod: Problem is that the oauth module is not working "out-of-the-box". That's why the config needs to be here :(

PR #103, opened 2019-09-04
WIP: Feature/cart item stock check

ChristopherSchwab:


category/domain/category.go
```diff
@@ -25,10 +25,18 @@ type (
...
 		CategoryTypeCode   string
 		CategoryAttributes Attributes
+		Promotion          Promotion
 	}
 
 	// Attributes define additional category attributes
 	Attributes map[string]interface{}
+
+	// Promotion defines promotion for a category
+	Promotion struct {
```
danielpoe: Why is this part of this PR?

cart/application/cartService.go
```diff
@@ -861,3 +861,33 @@ func (cs *CartService) dispatchAllEvents(ctx context.Context, events []flamingo.
 		cs.eventRouter.Dispatch(ctx, e)
 	}
 }
+
+// CheckAndAdjustStockInCart - checks if the stock of an item in the cart is lower than the quantity of it and if so updates the quantity
+func (cs *CartService) CheckAndAdjustStockInCart(ctx context.Context, session *web.Session) error {
```
danielpoe: Stock is a "concept" that was not known to the cart.
The cart has the concept of "QtyRestrictors"  and we should use this as naming.
e.g. AdjustToRestrictedQty

The result of that function should be a value object that descibes what has been adjusted (e.g. slice of AdjustmentInfo (cartid, old, new qty, restrictor?)

cart/interfaces/controller/cartcontroller.go
```diff
@@ -161,6 +167,11 @@ func (cc *CartViewController) UpdateQtyAndViewAction(ctx context.Context, r *web
 		}
 	}
 
+	err = cc.applicationCartService.CheckAndAdjustStockInCart(ctx, r.Session())
```
danielpoe: Imho we should enable this via feature flag - and per default it should be deactivated

cart/interfaces/controller/cartcontroller.go
```diff
@@ -126,6 +126,12 @@ func (cc *CartViewController) AddAndViewAction(ctx context.Context, r *web.Reque
 		return cc.responder.Render("checkout/carterror", nil)
 	}
 
+	err = cc.applicationCartService.CheckAndAdjustStockInCart(ctx, r.Session())
```
danielpoe: use featureflag here as well

category/domain/category.go
ChristopherSchwab: There was a mistake while rebasing. Will close this merge request and open a new one. Will also copy over your other comments.

PR #111, opened 2019-09-12
add possibility to configure default value for use billing address

zohmi:


zohmi:
@tessig I thought it is not necessary as it behaves as before. Only if you want to reverse the behaviour, you need to configure it. Does a default configuration in that case makes sense anyways?

tessig:
@zohmi It's a nice way to see all possible configuration keys of a module in one place

zohmi:
@tessig Ok, added the default config and updated the documentation

PR #112, opened 2019-09-16
price: return payable on SplitInPayables

masmrlar:
Price.SplitInPayables should return payable prices. Only creating new price from int with precision does not result into payable price values.

### DIFF
--- a/price/domain/price.go
+++ b/price/domain/price.go
@@ -325,14 +325,14 @@ func (p Price) GetPayableByRoundingMode(mode string, precision int) Price {
 	if p.IsNegative() {
 		negative = -1
 	}
-	offsetToCheckRounding := new(big.Float).Mul(p.precisionF(precision), new(big.Float).SetInt64(10))
+	offsetToCheckRounding := new(big.Float).Mul(p.precisionF(precision), new(big.Float).SetInt64(100))
 
 	amountTruncatedFloat, _ := new(big.Float).Mul(amountForRound, p.precisionF(precision)).Float64()
 	amountTruncatedInt := int64(amountTruncatedFloat)
 
 	amountRoundingCheckFloat, _ := new(big.Float).Mul(amountForRound, offsetToCheckRounding).Float64()
 	amountRoundingCheckInt := int64(amountRoundingCheckFloat)
-	valueAfterPrecision := (amountRoundingCheckInt - (amountTruncatedInt * 10)) * negative
+	valueAfterPrecision := (amountRoundingCheckInt - (amountTruncatedInt * 100)) * negative
 	if amountTruncatedFloat >= float64(math.MaxInt64) {
 		// will not work if we are already above MaxInt - so we return unrounded price:
 		newPrice.amount = p.amount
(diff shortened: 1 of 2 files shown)


price/domain/price.go
```diff
@@ -389,6 +389,11 @@ func (p Price) SplitInPayables(count int) ([]Price, error) {
 	if count <= 0 {
 		return nil, errors.New("Split must be higher than zero")
 	}
+
+	if count == 1 {
+		return []Price{p.Clone().GetPayable()}, nil
```
danielpoe: GetPAyable clone doch schon?

price/domain/price.go
```diff
@@ -419,7 +424,7 @@ func (p Price) SplitInPayables(count int) ([]Price, error) {
 		if p.IsNegative() {
 			splittedAmount *= -1
 		}
-		prices[i] = NewFromInt(splittedAmount, precision, p.Currency())
+		prices[i] = NewFromInt(splittedAmount, precision, p.Currency()).GetPayable()
```
danielpoe: Ich versteh nicht warum das notwendig ist? splittedAmount ist doch immer ein int und es dürfte keinen unterschied machen?

price/domain/price.go
masmrlar: Indeed - I'll remove that.

price/domain/price.go
masmrlar: The new price created by this NewFromInt does not create a payable.

You can try the test case "return payables".
Without this you get the correct sum up (which has been tested before only), but splitting up 88.39551 into 4 payables will result into 22.1, 22.1, 22.1, 22.09 if we don't apply GetPayable and return that.

As the payable representation for 88.39551 is 88.40 this imho fixes the issue.

We could also fix issues like this by calling GetPayable on every slice item we consume from the result of SplitInPayables.

Or - maybe a better solution:

we have a function Split which does not apply GetPayable on each item
and have the function SplitInPayables which has GetPayable applied on each item
to get it as flexible as possible.

danielpoe: I checked the testcase - its green also when I remove the "..GetPayable()" ? for me

But I think if its the case that NewFromInt does not return a Payable price in some circumstances we should make sure it does and add the .GetPayable() to NewFromInt() ?

masmrlar: Damn - I thought I pinned a combination where the error occurs. 

What do you think about having `Split` and `SplitInPayables` to have it as speaking as possible?

Maybe I also find the erratic calculation combination in my brain...

danielpoe: I think the whole reason of the Split is to get Payables - I see no reason in offering another Split
(then you can just do Quo with the Big Int)

masmrlar: mhm, indeed

price/domain/price.go
```diff
@@ -325,14 +325,14 @@ func (p Price) GetPayableByRoundingMode(mode string, precision int) Price {
 	if p.IsNegative() {
 		negative = -1
 	}
-	offsetToCheckRounding := new(big.Float).Mul(p.precisionF(precision), new(big.Float).SetInt64(10))
+	offsetToCheckRounding := new(big.Float).Mul(p.precisionF(precision), new(big.Float).SetInt64(100))
```
masmrlar: @danielpoe is it possible that we have in fact a too less precision on rounding?

I modified the GetPayable test and added a call for 88.39499 and got a Payable of 88.39 - but imho we should get a Payable of 88.40 for that value. 

Thus I increased the rounding precisions here and adapted the rounding up/down comparisons to 
- round up if we get xx.xx45 for halfup
- round down if we get xx.xx55 for halfdown

But I am also pretty unsure about this now...

masmrlar:
Nope, i am really completely wrong with this - closing the mr

PR #119, opened 2019-09-24
cart: Fix Cart Merging

danielpoe:
1) It is hard to test cartService because of Authentication and the problem to mock it => add Interfaces

2) Cart Merge currently has problems:
- No BillingAddress is merged
- AdditionalInfos are missing in Delivery
- its not handled as a "detached transaction" and a canceled context might kick in

### DIFF
--- a/cart/application/eventHandling.go
+++ b/cart/application/eventHandling.go
@@ -47,44 +47,73 @@ func (e *EventReceiver) Notify(ctx context.Context, event flamingo.Event) {
 		}
 	//Handle LoginEvent and Merge Cart
 	case *domain.LoginEvent:
-		if currentEvent == nil {
-			return
-		}
-		if !e.cartReceiverService.ShouldHaveGuestCart(currentEvent.Session) {
-			return
-		}
-		guestCart, err := e.cartReceiverService.ViewGuestCart(ctx, currentEvent.Session)
-		if err != nil {
-			e.logger.WithContext(ctx).Error("LoginEvent - Guestcart cannot be received %v", err)
-			return
-		}
-		if !e.cartReceiverService.IsLoggedIn(ctx, currentEvent.Session) {
-			e.logger.WithContext(ctx).Error("Received LoginEvent but user is not logged in!!!")
-			return
-		}
-		for _, d := range guestCart.Deliveries {
-			e.cartService.UpdateDeliveryInfo(ctx, currentEvent.Session, d.DeliveryInfo.Code, cartDomain.CreateDeliveryInfoUpdateCommand(d.DeliveryInfo))
-			for _, item := range d.Cartitems {
-				e.logger.WithContext(ctx).Debug("Merging item from guest to user cart %v", item)
-				addRequest := e.cartService.BuildAddRequest(ctx, item.MarketplaceCode, item.VariantMarketPlaceCode, item.Qty)
-				e.cartService.AddProduct(ctx, currentEvent.Session, d.DeliveryInfo.Code, addRequest)
+		web.RunWithDetachedContext(ctx, func(ctx context.Context) {
+			if currentEvent == nil {
+				return
 			}
-		}
+			if !e.cartReceiverService.ShouldHaveGuestCart(currentEvent.Session) {
+				return
+			}
+			guestCart, err := e.cartReceiverService.ViewGuestCart(ctx, currentEvent.Session)
+			if err != nil {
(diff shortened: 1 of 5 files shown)


carstendietrich:
Yay finally an AuthManagerInterface 👍 🎉 
Will have a look at the other changes later today.

danielpoe:
Only a local interface.. the generic one is part of the new auth module (see started branch in flamingo)

cart/domain/cart/cartServicePorts.go
```diff
@@ -98,10 +98,12 @@ var (
...
 //CreateDeliveryInfoUpdateCommand - factory to get the update command based on the given deliveryInfos (which might come from cart)
 func CreateDeliveryInfoUpdateCommand(info DeliveryInfo) DeliveryInfoUpdateCommand {
-	return DeliveryInfoUpdateCommand{
+	uc := DeliveryInfoUpdateCommand{
 		DeliveryInfo: info,
 		additional:   info.AdditionalDeliveryInfos,
 	}
+	uc.SetAdditional(info.AdditionalDeliveryInfos)
+	return uc
 }
```
jeinfeldt: I think we can remove line 103 as we always overwrite the value of additional in SetAdditional.

From a first glance it is a bit confusing why that value is assigned twice. 
First when initialising the struct and then again using the function. 

When checking out init I understood that we do not want that value to be nil.

jeinfeldt:
Regarding the failing travis pipeline, please run "go fmt" on the code!

cart/domain/cart/cartServicePorts.go
carstendietrich: I'll updated the code, it should now make more sense.

cart/application/eventHandling.go
```diff
@@ -47,44 +47,92 @@ func (e *EventReceiver) Notify(ctx context.Context, event flamingo.Event) {
...
+			}
+			if guestCart.HasAppliedCouponCode() {
+				for _, code := range guestCart.AppliedCouponCodes {
+					_, err := e.cartService.ApplyVoucher(ctx, currentEvent.Session, code.Code)
+					if err != nil {
+						e.logger.WithContext(ctx).Error("LoginEvent - customerCart ApplyVoucher has error", code.Code, err)
+					}
+				}
+			}
+			if guestCart.HasAppliedGiftCards() {
```
carstendietrich: @danielpoe GiftCards were missing, i'll added them. Please have a look.

cart/application/eventHandling.go
```diff
@@ -47,44 +47,92 @@ func (e *EventReceiver) Notify(ctx context.Context, event flamingo.Event) {
...
+			if !e.cartReceiverService.IsLoggedIn(ctx, currentEvent.Session) {
+				e.logger.WithContext(ctx).Error("Received LoginEvent but user is not logged in!!!")
+				return
+			}
+			customerCart, err := e.cartReceiverService.ViewCart(ctx, currentEvent.Session)
+			if err != nil {
+				e.logger.WithContext(ctx).Error("LoginEvent - customerCart cannot be received %v", err)
+				return
+			}
+			err = e.cartService.DeleteSavedSessionGuestCartID(currentEvent.Session)
```
carstendietrich: In the old implementation the session was deleted at the end, should we delete it at this point already?

cart/domain/cart/cartServicePorts.go
```diff
@@ -111,6 +113,17 @@ func (d *DeliveryInfoUpdateCommand) AddAdditional(key string, val AdditionalDeli
 	return err
 }
 
+// SetAdditional adds additional delivery info data
+func (d *DeliveryInfoUpdateCommand) SetAdditional(val map[string]json.RawMessage) {
+	d.init()
+
+	if val == nil {
```
carstendietrich: @danielpoe added check for nilness.

carstendietrich:
> Regarding the failing travis pipeline, please run "go fmt" on the code!

did it while adding the misssing giftcard stuff.

cart/application/eventHandling.go
danielpoe: That was by intention - i thought it makes sense to "remove" guest cart right after its clear we have a customer cart.
Otherwise it might take n API calls time to delete it - do you see a drawback?

carstendietrich: Ah okay if its intended that's fine. I just wasn't sure if we should keep the guest cart intact in case of an later error.

PR #122, opened 2019-09-27
cart: Fix validators/restrictors not able to handle configurable product

carstendietrich:
When working with products we always need to check if its an configurable and get the active variant to be able to restrict and validate the variant instead of the base product.


cart/application/cartService.go
```diff
@@ -896,6 +911,16 @@ func (cs *CartService) AdjustItemsToRestrictedQty(ctx context.Context, session *
 				return nil, err
 			}
 
+			if product.Type() == productDomain.TypeConfigurable {
```
ChristopherSchwab: Smells like duplicate code. Maybe move to separate function?

carstendietrich: Moved to a separate function, please have a look.

PR #127, opened 2019-10-04
product: add possibility to provide labels for product attribute codes

tessig:


product/interfaces/controller/controller_test.go
```diff
@@ -32,342 +90,371 @@ func (mps *MockProductService) Get(ctx context.Context, marketplacecode string)
...
+		"name":            "my-product-title",
+	}
 
+	result := vc.Get(ctx, r)
+	require.IsType(t, &web.RenderResponse{}, result)
+
+	renderResponse := result.(*web.RenderResponse)
+	assert.Equal(t, "product/product", renderResponse.Template)
+	require.IsType(t, productViewData{}, renderResponse.DataResponse.Data)
+	p, _ := new(MockProductService).Get(nil, "simple")
```
golangcibot: SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (from `staticcheck`)

PR #136, opened 2019-10-17
search: Make searchService implementation optional 

danielpoe:
Background:
- interface layer of product and category package depend on search package - e.g. they reuse dto and graphQL types
- it should be possible to reuse them without a searchservice adapter. Currently thats not possible since in Module the controller and its dependencies request a Searchservice

=> We should make this DI optional

### DIFF
--- a/search/application/searchService.go
+++ b/search/application/searchService.go
@@ -6,6 +6,7 @@ package application
 
 import (
 	"context"
+	"errors"
 	"net/url"
 
 	"flamingo.me/flamingo-commerce/v3/search/domain"
@@ -63,6 +64,9 @@ func (s *SearchService) Inject(
 
 // FindBy returns a SearchResult for the given Request
 func (s *SearchService) FindBy(ctx context.Context, documentType string, searchRequest SearchRequest) (*SearchResult, error) {
+	if s.searchService == nil {
+		return nil, errors.New("No searchservice available")
+	}
 	var currentURL *url.URL
 	request := web.RequestFromContext(ctx)
 	if request == nil {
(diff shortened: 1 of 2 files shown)


search/application/searchService.go
```diff
@@ -55,16 +72,16 @@ func (s *SearchService) FindBy(ctx context.Context, documentType string, searchR
...
 
 	// pageSize can either be set in the request, or we use the configured default or if nothing set we rely on the ProductSearchService later
 	pageSize := searchRequest.PageSize
 	if pageSize == 0 {
-		pageSize = int(s.DefaultPageSize)
+		pageSize = int(s.defaultPageSize)
 	}
 
-	result, err := s.SearchService.SearchFor(ctx, documentType, BuildFilters(searchRequest, pageSize)...)
+	result, err := s.searchService.SearchFor(ctx, documentType, BuildFilters(searchRequest, pageSize)...)
```
tessig: since searchService is optional now, we should insert guard clauses to prevent nil pointer dereferences

search/application/searchService.go
```diff
@@ -104,16 +121,16 @@ func (s *SearchService) Find(ctx context.Context, searchRequest SearchRequest) (
...
 
 	// pageSize can either be set in the request, or we use the configured default or if nothing set we rely on the ProductSearchService later
 	pageSize := searchRequest.PageSize
 	if pageSize == 0 {
-		pageSize = int(s.DefaultPageSize)
+		pageSize = int(s.defaultPageSize)
 	}
 
-	result, err := s.SearchService.Search(ctx, BuildFilters(searchRequest, pageSize)...)
+	result, err := s.searchService.Search(ctx, BuildFilters(searchRequest, pageSize)...)
```
tessig: also here a guard clause

search/application/searchService.go
danielpoe: Thanks make sense - added guard

PR #147, opened 2019-11-01
Make toString function for attributes

dylan-provost:
A proper attribute struct was introduce recently for categories. To make the values more accessible we added two "toString" functions to return a concatenated string of all the values of an attribute, and all the values of an attribute value.

### DIFF
--- a/category/domain/category.go
+++ b/category/domain/category.go
@@ -129,13 +129,23 @@ func (a Attributes) All() []Attribute {
 	return att
 }
 
+// ToString returns a concatenated string of all the values of all attributes
+func (a Attributes) ToString() string {
+	var attValue []string
+	for _, v := range a {
+		attValue = append(attValue, v.ToString())
+	}
+
+	return strings.Join(attValue[:], ",")
+}
+
 // ToString returns a concatenated string of all the values of an attribute
 func (a Attribute) ToString() string {
 	var attValue []string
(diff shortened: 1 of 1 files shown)


category/domain/category.go
```diff
@@ -128,6 +129,25 @@ func (a Attributes) All() []Attribute {
...
+	return strings.Join(attValue[:], ",")
+}
+
+// ToString returns a concatenated string of all the values of an attribute
+func (a Attribute) ToString() string {
+	var attValue []string
+	for _, v := range a.Values {
+		attValue = append(attValue, v.Value())
+	}
+	return strings.Join(attValue[:], ",")
```
danielpoe: The [:] is not required . just attValue should be enough.

category/domain/category.go
```diff
@@ -128,6 +129,25 @@ func (a Attributes) All() []Attribute {
 	return att
 }
 
+// ToString returns a concatenated string of all the values of all attributes
+func (a Attributes) ToString() string {
```
danielpoe: Do we really need a function that returns all Attributes Values as a list?
e.g. you have:

colour: red,green
size: xl,L

Ans then it would return a string "red,green,xl,l"
for what?

dylan-provost: Good point, team requested that, so I did. Just discussed through it, and there is no strong reason why, I remove it

PR #150, opened 2019-11-12
Provide Search GraphQL Interfaces

lifedraft:


search/interfaces/graphql/schema.graphql
```diff
@@ -17,4 +17,25 @@ type Commerce_Search_Meta {
...
+    selectedDesc: Boolean!
+}
+
+type Commerce_Search_Suggestion {
+    text:      String!
+    highlight: String!
+}
+
+type Commerce_Search_Result {
+    # Hits: []Document
```
danielpoe: why not adding hits right now? Or if we dont need Commerce_Search_Result we can uncomment it completely?

search/interfaces/graphql/schema.graphql
```diff
@@ -17,4 +17,25 @@ type Commerce_Search_Meta {
...
+    searchMeta:     Commerce_Search_Meta!
+    # Facets: domain.FacetCollection
+    suggestions:    [Commerce_Search_Suggestion]
```
danielpoe: in Commerce_Product_SearchResut it says " [Commerce_Search_Suggestion!] " - whith or without "!" ?

product/interfaces/graphql/resolver.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	return &application.SearchResult{
+		Suggestions: result.Suggestion,
+		Products:    result.Hits,
+		SearchMeta:  result.SearchMeta,
+		Facets:      result.Facets,
+	}, nil
+
+	return nil, nil
```
golangcibot: unreachable: unreachable code (from `govet`)

search/interfaces/graphql/dto/dto.go
```diff
@@ -7,6 +7,7 @@ type CommerceSearchRequest struct {
 	SortBy          string
 	SortDirection   string
 	KeyValueFilters []CommerceSearchKeyValueFilter
+	Query           *string
```
jeinfeldt: Just out of curiosity, any specific reason why a pointer is used here? To enable nil setting?

kalypso-byte: Good catch. I think it was done because of that, but it's not necessary because the search does not expect a pointer anyway.

danielpoe: I think it was because the graphql lib suggested this - if graphql model mapping works without pointer then lets change this.

kalypso-byte: Ah okay! I checked if it is working without the pointer and seems like it does :)

search/domain/service.go
```diff
@@ -36,13 +36,20 @@ type (
...
-		SelectedAsc  bool
+		// Label that you normally want to show in the frontend (e.g. "Price")
+		Label string
+		// Field that you need to use in SearchRequest>SortFilter
+		Field string
+		// SelectedAsc true if sorting by this field is active
+		SelectedAsc bool
+		// SelectedDesc true if sorting by this field is active
 		SelectedDesc bool
+		// Asc - DEPRECATED - used to give the "Field" that is used to trigger ascending search
```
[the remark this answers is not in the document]
kalypso-byte: I updated the deprecated fields. Which fields do you mean? Field or Asc/Desc? We left Asc/Desc there because we weren't sure if we introduce a breaking change if we remove them completely. In case of Field. We are using it for the graphql part and it is used in flamingo-om3 which is in the other merge request.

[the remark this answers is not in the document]
panzerfahrer: @danielpoe could both fields ( `Asc`, `Desc`) actually be deleted? this looks quite like dead code.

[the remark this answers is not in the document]
danielpoe: How do you know they are not used??
They are part of the public API - projects depend on it and its also filled in the Adapters currently
@panzerfahrer

search/domain/service.go
```diff
@@ -36,13 +36,22 @@ type (
...
+		// Label that you normally want to show in the frontend (e.g. "Price")
+		Label string
+		// Field that you need to use in SearchRequest>SortFilter
+		Field string
+		// SelectedAsc true if sorting by this field is active
+		SelectedAsc bool
+		// SelectedDesc true if sorting by this field is active
 		SelectedDesc bool
+		// Asc - represents the field that is used to trigger ascending search.
+		// Deprecated: use "Field" and "SelectedAsc" instead to set which field should be sortable
```
danielpoe: The message is wrong
It should say: Use "Field" to know which Field to set in the SortFilter

search/domain/service.go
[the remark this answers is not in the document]
panzerfahrer: The team decided to deprecate the field. If the fields are actually in use, we should probably add fallback behavior. Or do we need a rollback with another solution?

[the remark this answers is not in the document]
panzerfahrer: ... @kalypso-byte could you please clarify, whether the deprecation was a valid move.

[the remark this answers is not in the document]
danielpoe: The DEPRECATED is valid  we did this together. - but because the intent of this property is wrong (it represents the fieldname for searchperience)
We just need time to adjust the current usage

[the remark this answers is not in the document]
danielpoe: So all is fine at the moment - just wanted to clarify

PR #156, opened 2019-11-25
cart + price: Add custom attributes to charge

ChristopherSchwab:


danielpoe:
The Pull Request describes no purpose why a Charge would need this?
@ChristopherSchwab

ChristopherSchwab:
> The Pull Request describes no purpose why a Charge would need this?
> @ChristopherSchwab

We implemented these custom attributes because we had a usecase that required charges to be of the same type but still be differentiable internally. In our concrete case we have to distinguish if a gift card charge is handled internally or by an extern service, which we note in a custom attribute. 

@danielpoe

PR #159, opened 2019-11-27
Revert charge custom attributes and add gift card filter function

ChristopherSchwab:
Reverts the changes from this PR: https://github.com/i-love-flamingo/flamingo-commerce/pull/156

Also adds a function to filter a single gift card from a slice of gift cards by its code.

### DIFF
--- a/cart/domain/cart/cart.go
+++ b/cart/domain/cart/cart.go
@@ -50,7 +50,7 @@ type (
 		Totalitems []Totalitem
 
 		//List of applied gift cards
-		AppliedGiftCards []AppliedGiftCard
+		AppliedGiftCards AppliedGiftCards
 	}
 
 	// Teaser - represents some teaser infos for cart
--- a/cart/domain/cart/cart.go
+++ b/cart/domain/cart/cart.go
@@ -50,7 +50,7 @@ type (
 		Totalitems []Totalitem
 
 		//List of applied gift cards
-		AppliedGiftCards AppliedGiftCards
+		AppliedGiftCards []AppliedGiftCard
 	}
 
 	// Teaser - represents some teaser infos for cart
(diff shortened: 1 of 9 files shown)


cart/domain/cart/giftcard.go
```diff
@@ -99,3 +99,14 @@ func (cards *AppliedGiftCards) ByRemaining() AppliedGiftCards {
 	}
 	return result
 }
+
+// GiftCardByCode returns a single gift card if the given code matches its code.
+func (cards *AppliedGiftCards) GiftCardByCode(code string) (card AppliedGiftCard, found bool) {
+	for _, card = range *cards {
```
carstendietrich: maybe add a filter function like in AppliedDiscounts?

cart/domain/cart/cart.go
```diff
@@ -50,7 +50,7 @@ type (
 		Totalitems []Totalitem
 
 		//List of applied gift cards
-		AppliedGiftCards []AppliedGiftCard
+		AppliedGiftCards AppliedGiftCards
```
carstendietrich: This is a potential breaking change we should try to avoid this.

cart/domain/cart/giftcard.go
```diff
@@ -99,3 +99,14 @@ func (cards *AppliedGiftCards) ByRemaining() AppliedGiftCards {
...
+		if card.Code == code {
+			return card, true
+		}
+	}
+
+	return AppliedGiftCard{}, false
```
danielpoe: why not nil, false (change to pointer) 
I know that we said that dealing with Values objects over References - but imho thats more relevant for parameter in called methods.
In case something is optional I find its more logic to return nil instead an unfilled struct of that type. We do it at other places as well this way

ChristopherSchwab: Yeah, that makes sense. See: https://github.com/i-love-flamingo/flamingo-commerce/pull/160

PR #160, opened 2019-11-28
cart: Change return type of GiftCardByCode to pointer

ChristopherSchwab:


cart/domain/cart/giftcard.go
```diff
@@ -101,12 +101,11 @@ func (cards *AppliedGiftCards) ByRemaining() AppliedGiftCards {
 }
 
 // GiftCardByCode returns a single gift card if the given code matches its code.
-func (cards *AppliedGiftCards) GiftCardByCode(code string) (card AppliedGiftCard, found bool) {
-	for _, card = range *cards {
+func (cards *AppliedGiftCards) GiftCardByCode(code string) (card *AppliedGiftCard, found bool) {
+	for _, card := range *cards {
```
ChristopherSchwab: The `card` var of the return parameters is now overshadowed by this `card` var. This is because it is not possible to assign to a pointer via range. Should we remove the named return parameters?

PR #164, opened 2019-12-11
cart: Add idempotency key proof of concept for payment selection

carstendietrich:
Proof of concept for adding an idempotency key to the payment selection to use it later during communication with the payment gateway / provider.

**ToDo**
Update Checkout Controller: If payment gateway response with Error during payment we need to generate a new key (cart.PaymentSelecetion.GenerateNew..) to allow user retries.


carstendietrich:
work will continue in #171

PR #165, opened 2019-12-12
cart/checkout: Reduce calls for updating cart items

zohmi:


cart/domain/cart/cartServicePorts.go
```diff
@@ -41,6 +41,7 @@ type (
 	ModifyBehaviour interface {
 		DeleteItem(ctx context.Context, cart *Cart, itemID string, deliveryCode string) (*Cart, DeferEvents, error)
 		UpdateItem(ctx context.Context, cart *Cart, itemID string, deliveryCode string, itemUpdateCommand ItemUpdateCommand) (*Cart, DeferEvents, error)
```
danielpoe: Maybe we should think of adding all relevant properties to the ItemUpdateCommand:
- itemId
- deliveryCode

Then the UpdateItems can get a slice of UpdateCommands

cart/infrastructure/InMemoryBehaviour.go
```diff
@@ -138,6 +138,56 @@ func (cob *InMemoryBehaviour) UpdateItem(ctx context.Context, cart *domaincart.C
 	return cob.resetPaymentSelectionIfInvalid(ctx, cart)
 }
 
+// UpdateItems updates multiple cart items
+func (cob *InMemoryBehaviour) UpdateItems(ctx context.Context, cart *domaincart.Cart, deliveryCode string, itemUpdateCommands domaincart.ItemUpdateCommands) (*domaincart.Cart, domaincart.DeferEvents, error) {
+	if !cob.cartStorage.HasCart(cart.ID) {
+		return nil, nil, fmt.Errorf("cart.infrastructure.InMemoryBehaviour: Cannot update - Guestcart with id %v not existent", cart.ID)
+	}
+	if delivery, ok := cart.GetDeliveryByCode(deliveryCode); ok {
```
danielpoe: we should reuse cob.UpdateItem() in the loop - no need to duplicate code

checkout/domain/sourcingengine.go
```diff
@@ -112,18 +113,23 @@ func (se *SourcingEngine) SetSourcesForCartItems(ctx context.Context, session *w
...
 			se.Logger.WithContext(ctx).WithField("category", "checkout").WithField("subcategory", "SourcingEngine").Debug("SourcingEngine detected source %v for item %v", sourceID, decoratedCartItem.Item.ID)
-			err = se.Cartservice.UpdateItemSourceID(ctx, session, decoratedCartItem.Item.ID, decoratedDelivery.Delivery.DeliveryInfo.Code, sourceID)
-			if err != nil {
-				return errors.Wrap(err, "Could not update cart item")
-			}
+
+			itemSources[decoratedCartItem.Item.ID] = sourceID
+		}
+
+		err := se.Cartservice.UpdateItemsSourceID(ctx, session, decoratedDelivery.Delivery.DeliveryInfo.Code, itemSources)
```
danielpoe: If change suggested above is done we can collect all update commands and just have one call

cart/domain/cart/cartServicePorts.go
```diff
@@ -77,9 +78,16 @@ type (
 		// Source Id of where the items should be initial picked - This is set by the SourcingLogic
 		SourceID       *string
 		Qty            *int
+		ItemID         string
```
carstendietrich: We should at least add a comment here that ItemID and DeliveryCode are used for finding the item in the cart and therefore are mandatory. the fields are not for changing the items for example the delivery code of item xy

cart/application/cartService.go
```diff
@@ -337,10 +340,36 @@ func (cs *CartService) UpdateItemSourceID(ctx context.Context, session *web.Sess
...
+		cs.logger.WithContext(ctx).WithField("subCategory", "UpdateItemSourceId").Error(err)
+
+		return err
+	}
+
+	return nil
+}
+
+// UpdateItemsSourceID updates multiple item source ids
+func (cs *CartService) UpdateItemsSourceID(ctx context.Context, session *web.Session, updateCommands cartDomain.ItemUpdateCommands) error {
```
carstendietrich: The function name is misleading since you could also update the qty with this call depending on the content of the `updateCommands`.. Maybe we should go for `UpdateItems`

cart/infrastructure/InMemoryBehaviour.go
```diff
@@ -99,23 +99,66 @@ func (cob *InMemoryBehaviour) DeleteItem(ctx context.Context, cart *domaincart.C
...
+
+	err := cob.cartStorage.StoreCart(cart)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "cart.infrastructure.InMemoryBehaviour: error on saving cart")
+	}
+
+	return cob.resetPaymentSelectionIfInvalid(ctx, cart)
+}
+
+func (cob *InMemoryBehaviour) updateItem(ctx context.Context, cart *domaincart.Cart, itemUpdateCommand domaincart.ItemUpdateCommand) error {
```
carstendietrich: Should we add error handling for delivery for supplied delivery not found? or supplied item id not in delivery? We have vars for those cases in cartServicePort: ErrItemNotFound, ErrDeliveryCodeNotFound etc

cart/application/cartService.go
carstendietrich: Check for mandatory fields om ItemUpdateCommands (ItemID, deliveryCode) could potentially be added here

cart/domain/cart/cartServicePorts.go
danielpoe: Since we anyhow change the interface..

Didn't we decide a while ago that itemId is the unique idendifier of an item and should be unique for complete cart?

carstendietrich: True maybe we then just use the ItemID here. Maybe we add a function to get the delivery of an item in case that the underlying adapter needs the deliverycode

carstendietrich: Something like cart.GetDeliveryByItem(..)

PR #170, opened 2020-01-07
only validate date of birth in personal data form if it is required

SchiffFlieger:


cart/interfaces/controller/forms/personaldataform.go
```diff
@@ -156,17 +156,15 @@ func (p *DefaultPersonalDataFormService) Validate(ctx context.Context, req *web.
...
-	}
-
-	if p.minAge > 0 {
-		if !validateMinimumAge(personalDataForm.DateOfBirth, p.minAge) {
-			validationInfo.AddFieldError("personalData.dateOfBirth", "formerror_dateOfBirth_tooyoung", "you need to be at least "+strconv.Itoa(p.minAge))
+	if p.dateOfBirthRequired {
+		if personalDataForm.DateOfBirth == "" {
+			validationInfo.AddFieldError("personalData.dateOfBirth", "formerror_dateOfBirth_required", "date of birth is required")
+		} else if !validateDOB(personalDataForm.DateOfBirth) {
+			validationInfo.AddFieldError("personalData.dateOfBirth", "formerror_dateOfBirth_formaterror", "date of birth has wrong format required: yyyy-mm-dd")
```
danielpoe: At least this makes sense to validate even if not required?
(In case an empty field does not trigger validation error)

SchiffFlieger: Yes, you are right. The validation failed if the date was empty and not required. But of course we still should validate it, if it is not empty.

PR #173, opened 2020-01-07
Extend GraphQL schema for Cart (Sum*, Giftcard, Discounts) and Price charges

lifedraft:


danielpoe:
> _No description provided._

Can you please provide a desrciption of problem / requirement?

cart/domain/cart/cart.go
```diff
@@ -461,6 +479,15 @@ func (c Cart) SumTotalDiscountAmount() domain.Price {
 	return price
 }
 
+// SumTotalDiscountWithGiftCardsAmount – returns sum price of total discounts with applied gift cards
+func (c Cart) SumTotalDiscountWithGiftCardsAmount() domain.Price {
```
danielpoe: Semantically I find it hard to get the intent:

Discount is something that reduced the price that need to be payed.

Giftcard is a way of paying something.

Why would we provide this in the cart domain?

If for display reasons you want to sum that up - you are free to do it in the interface layer..

Imho hard to argument that it belongs to a typical thing you do in the cart domain (summing a discount and a gifcart payment)

cart/domain/cart/cart.go
```diff
@@ -601,6 +628,30 @@ func (t Taxes) TotalAmount() domain.Price {
 	return result
 }
 
+// GetAllTaxes – returns all Taxes as slice
+func (t Taxes) GetAllTaxes() []Tax {
```
danielpoe: Taxes is already a slice of Tax:

Taxes []Tax

you can range over the variable taxes itself - no need for this Getter

cart/domain/cart/cart.go
```diff
@@ -601,6 +628,30 @@ func (t Taxes) TotalAmount() domain.Price {
...
+
+	if len(taxes) > 0 {
+		return taxes
+	}
+
+	return nil
+}
+
+// GetByType - returns tax by given type
+func (t Taxes) GetByType(taxType string) Tax {
```
danielpoe: Because Tax can not exist for a type imho makes sense to return pointer and nil in case its not found

cart/domain/cart/cart.go
```diff
@@ -750,12 +801,54 @@ func (p PricedItems) TotalItems() map[string]domain.Price {
 	return p.totalItems
 }
 
-//ShippingItems - returns the Price per ShippingItems - map key is deliverycode
+// GetAllTotalItems – return all total items
+func (p PricedItems) GetAllTotalItems() []PricedTotalItem {
```
danielpoe: Duplicate with function TotalItems

we should decide on one of the options

cart/interfaces/graphql/schema.graphql
```diff
@@ -17,15 +17,98 @@ type Commerce_Cart {
...
+    sumItemRelatedDiscountAmount: Commerce_Price!
+    hasAppliedCouponCode: Boolean!
+
+    getPaymentReference: String!
+    getTotalItemsByType(typeCode: String!): [Commerce_CartTotalitem!]
+    grandTotalCharges: Commerce_Price_Charges!
+
+    hasAppliedGiftCards: Boolean!
+    hasRemainingGiftCards: Boolean!
+    sumAppliedGiftCardsForced: Commerce_Price!
```
danielpoe: Forced in the shema sound like a not so nice property name?

price/interfaces/graphql/schema.graphql
```diff
@@ -13,4 +13,17 @@ type Commerce_Charge {
...
+input Commerce_ChargeQualifierInput {
+    type: String!
+    reference: String!
+}
+
+type Commerce_Price_Charges {
+    items: [Commerce_Charge!]
+    hasType(ctype: String): Boolean
+    hasChargeQualifier(qualifier: Commerce_ChargeQualifierInput!): Boolean
+    getByChargeQualifierForced(qualifier: Commerce_ChargeQualifierInput!): Commerce_Charge
```
danielpoe: Why can we not reuse the type Commerce_ChargeQualifier ? Would make sense to me - it is meant as a "key" that qualifies one Charge in a list of Charges

lifedraft:
> > _No description provided._
> 
> Can you please provide a desrciption of problem / requirement?

We are currently working on the cart summary element where we would show the totals, taxes, applied coupons and applied gift cards. Therefor we need this information in the graphql.

cart/domain/cart/cart.go
lifedraft: From concept side we thought we would sum up all "discounts" (cart rules, coupons and gift cards) and display the "discounts" to customer. Yes this is probably something very specific, in the old days we could've done this in the template, but in the graphql "world" this should be in the interface layer.

cart/domain/cart/cart.go
lifedraft: This one is tricky in graphql, truly Taxes could be a slice of Taxes but then it's not possible to access getter methods of Taxes type like the TotalAmount or GetByType. Maybe i'm missing something?

lifedraft: I'm also unsure about the get all methods, in other places we have sub property for items, like the PricedItems.

cart/interfaces/graphql/schema.graphql
lifedraft: Yup, we could remap this in field in graphql service. What do you think about general convention with the Forced suffix? 

Pattern i've found is methods with multiple returns are wrapped by the ...Forced  methods, like for example ```.ChargesByType().GetByTypeForced(...)```

price/interfaces/graphql/schema.graphql
lifedraft: In graphql input types follow an naming convention like in this case you have to suffix your input type with "...Input"

Therefore I've tried to prevent a breaking change because the Commerce_ChargeQualifier type was already introduced but never used.

cart/domain/cart/cart.go
lifedraft: Party true, TotalItems return map[string] of Price which unfortunately can't be used in graphql. I'll move this one into interface layer.

PR #174, opened 2020-01-09
Extend data layer

dylan-provost:
- Improve linting
- Add language setter
- Fix logs


dylan-provost:
> why did this not work?

Actually the locale.locale works perfectly. This is just because there are use cases were the page language isn't the same as the website's locale ie. having only some pages translated.

PR #175, opened 2020-01-11
cart: increase test coverage for cart

sebastian-keller:


carstendietrich:
Thanks for contributing! 🎉

PR #180, opened 2020-01-17
cart: Add GiftCard convenience function Total()

ChristopherSchwab:


cart/domain/cart/giftcard.go
```diff
@@ -35,6 +35,15 @@ func (card AppliedGiftCard) HasRemaining() bool {
 	return !card.Remaining.IsZero()
 }
 
+// Total returns the total value of the gift card by adding what is applied and remaining
```
jeinfeldt: Please also describe the error case as this function still returns the value of remaining in case the values cannot be added
```suggestion
// Total returns the total value of the gift card by adding what is applied and remaining
// In case the values cannot be added the function returns the remaining amount of the giftcard and an error
```

PR #183, opened 2020-01-23
cart: fix GraphQL for cart, due to new type alias in cart model

carstendietrich:
A new type alias for applied coupons was introduced in the cart model, this messed up the GraphQL mapping. This commit reverts the type alias while maintaining the needed behaviour in cartService.go

### DIFF
--- a/cart/application/cartService.go
+++ b/cart/application/cartService.go
@@ -970,7 +970,8 @@ func (cs *CartService) AdjustItemsToRestrictedQty(ctx context.Context, session *
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


carstendietrich:
@kalypso-byte could you please have a look if this fixes your graphQL issues?
@ChristopherSchwab this should keep the intended behaviour.. right?

ChristopherSchwab:
> @kalypso-byte could you please have a look if this fixes your graphQL issues?
> @ChristopherSchwab this should keep the intended behaviour.. right?

Yes, this keeps the intended behaviour :+1:

Issue #184, opened 2020-01-27
Behaviour of ProductCount() in cart.go should be unique


tlayh:
Just wondering if the method  ProductCount() has the correct behaviour. If there is the same product in two deliveries it will be counted twice. Shouldn't this function return the unique product count over all deliveries?

https://github.com/i-love-flamingo/flamingo-commerce/blob/11a1fcd1cf820b81a7ec6f7e986d393f3cba40cb/cart/domain/cart/cart.go#L280

PR #185, opened 2020-01-28
Behaviour of ProductCount() in cart.go should be unique #184

tlayh:
#184 will be solved with this pull request. Old function stays the same, but there is a ProductCountUnique now, that will check across deliveries.

### DIFF
--- a/cart/domain/cart/cart.go
+++ b/cart/domain/cart/cart.go
@@ -276,7 +276,7 @@ func (c Cart) ItemCount() int {
 	return count
 }
 
-// ProductCount - returns amount of different products
+// ProductCount returns the amount of different products, but a product is counted twice if it is in different deliveries
 func (c Cart) ProductCount() int {
 	count := 0
 	for _, delivery := range c.Deliveries {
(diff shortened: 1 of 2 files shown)


cart/domain/cart/cart.go
```diff
@@ -286,6 +286,19 @@ func (c Cart) ProductCount() int {
 	return count
 }
 
+// ProductCountUnique returns the amount of unique products across all deliveries
+func (c Cart) ProductCountUnique() int {
+	marketplaceCodes := make(map[string]string)
```
carstendietrich: you could use `make(map[string]struct{})`, since an empty struct doesn't take up memory. 

```diff
- marketplaceCodes := make(map[string]string)
+ marketplaceCodes := make(map[string]struct{})
```

```diff
- marketplaceCodes[item.MarketplaceCode] = item.MarketplaceCode
+ marketplaceCodes[item.MarketplaceCode] = struct{}{}
```

tlayh: RThx for the hint. Learned something new. Added the suggested improvment.

PR #186, opened 2020-01-30
cart: Add additonalData to AddRequest to be able to provide some when adding a product

ChristopherSchwab:
Similar to being able to put some additonal data when updating an item this would enable one to provide some when adding a product.


ChristopherSchwab:
> Looks good so far, could please add the change to the changelog?

Added to changelog

PR #187, opened 2020-02-06
cart: Adjust behaviour of add event and update qty event

jeinfeldt:
Hello friends,

currently the `AddToCartEvent` and `ChangedQtyInCartEvent` are published before the cart
has been updated in the cache as they are not part of the deferred events.

When listening to the event as a client, I want to receive the updated cart for evaluation.
This PR addresses this issue by adding the cart to said events, analogue to the `OrderPlacedEvent `.

I think this is how this was intended anyway, as the function `publishAddtoCartEvent` already receives the cart, but it is not used.

Another issue is that the `ChangedQtyInCartEvent` is thrown before the Cart is even updated, which means that the event is thrown even if there is an error with the update.

I am aware that this is a breaking change, but I think the behaviour so far is not correct.

Please review!

### DIFF
--- a/cart/Changelog.md
+++ b/cart/Changelog.md
@@ -60,6 +60,7 @@
 include a cart as a parameter
 * Breaking Change to behaviour of `AddToCartEvent` and `ChangedQtyInCartEvent`, they are now thrown after
 the cart has been adjusted and written back to cache
+* Events deferred from `ModifyBehaviour` are dispatched before `AddToCartEvent` and `ChangedQtyInCartEvent`
 * The `AddToCartEvent` includes the current cart (with added product)
 * The `ChangedQtyInCartEvent` includes the current cart (with updated quantities)
 * Add Whitebox Test `TestCartService_CartInEvent` to check `AddToCartEvent`
--- a/cart/Changelog.md
+++ b/cart/Changelog.md
@@ -58,6 +58,8 @@
 # 6. February 2020
 * Breaking Change to `EventPublisher` interface, `PublishChangedQtyInCartEvent` and `PublishAddToCartEvent` now
 include a cart as a parameter
+* Breaking Change to behaviour of `AddToCartEvent` and `ChangedQtyInCartEvent`, they are now thrown after
+the cart has been adjusted and written back to cache
 * The `AddToCartEvent` includes the current cart (with added product)
 * The `ChangedQtyInCartEvent` includes the current cart (with updated quantities)
 * Add Whitebox Test `TestCartService_CartInEvent` to check `AddToCartEvent`
--- a/cart/Changelog.md
+++ b/cart/Changelog.md
@@ -54,3 +54,10 @@
 
 # 17. January 2020
 * Add `AppliedGiftCard` convenience function `Total()`
+
+# 6. February 2020
+* Breaking Change to `EventPublisher` interface, `PublishChangedQtyInCartEvent` and `PublishAddToCartEvent` now
+include a cart as a parameter
+* The `AddToCartEvent` includes the current cart (with added product)
+* The `ChangedQtyInCartEvent` includes the current cart (with updated quantities)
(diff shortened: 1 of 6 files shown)


cart/application/cartService_test.go
```diff
@@ -931,3 +933,88 @@ func (mpos *MockPlaceOrderService) CancelGuestOrder(ctx context.Context, orderIn
...
+	)
+
+	// add product to cart, we expect event to be thrown with updated cart
+	addRequest := cartDomain.AddRequest{
+		MarketplaceCode: "code-1",
+		Qty:             1,
+	}
+	ctx := context.Background()
+	session := web.EmptySession()
+	cartService.AddProduct(ctx, session, "default_delivery_code", addRequest)
```
golangcibot: Error return value of `cartService.AddProduct` is not checked (from `errcheck`)

cart/application/cartService.go
```diff
@@ -298,7 +298,7 @@ func (cs *CartService) UpdateItemQty(ctx context.Context, session *web.Session,
 		return err
 	}
 
-	cs.eventPublisher.PublishChangedQtyInCartEvent(ctx, item, qtyBefore, qty, cart.ID)
+	cs.eventPublisher.PublishChangedQtyInCartEvent(ctx, cart, item, qtyBefore, qty)
```
ChristopherSchwab: Since the event is published before the `behaviourUpdateItem` call the cart will not be updated yet.

jeinfeldt: You are right, updated implementation

cart/application/cartReceiver_test.go
```diff
@@ -214,6 +221,24 @@ func (m *MockUserService) IsLoggedIn(ctx context.Context, session *web.Session)
...
+		mock.Mock
+	}
+)
+
+var _ flamingo.EventRouter = new(MockEventRouter)
+
+func (m *MockEventRouter) Dispatch(ctx context.Context, event flamingo.Event) {
+	// we just write the event type and the marketplace code to the mock, so we don't have to compare
+	// the complete cart
+	switch event.(type) {
```
golangcibot: S1034: assigning the result of this type assertion to a variable (switch event := event.(type)) could eliminate the following type assertions:
	cart/application/cartReceiver_test.go:237:19 (from `gosimple`)

cart/application/cartService.go
```diff
@@ -749,12 +815,6 @@ func (cs *CartService) checkProductQtyRestrictions(ctx context.Context, product
 	return nil
 }
 
-func (cs *CartService) publishAddtoCartEvent(ctx context.Context, currentCart cartDomain.Cart, addRequest cartDomain.AddRequest) {
-	if cs.eventPublisher != nil {
-		cs.eventPublisher.PublishAddToCartEvent(ctx, addRequest.MarketplaceCode, addRequest.VariantMarketplaceCode, addRequest.Qty)
-	}
-}
-
```
ChristopherSchwab: The interface function is now not used anymore within the repo. Maybe we can/should remove it entirely?

carstendietrich: good point unused unexported functions can safely be removed :)

cart/Changelog.md
```diff
@@ -53,4 +53,13 @@
...
 # 17. January 2020
-* Add `AppliedGiftCard` convenience function `Total()`
\ No newline at end of file
+* Add `AppliedGiftCard` convenience function `Total()`
+
+# 6. February 2020
+* Breaking Change to `EventPublisher` interface, `PublishChangedQtyInCartEvent` and `PublishAddToCartEvent` now
+include a cart as a parameter
+* Breaking Change to behaviour of `AddToCartEvent` and `ChangedQtyInCartEvent`, they are now thrown after
+the cart has been adjusted and written back to cache
```
carstendietrich: Please mention the new dispatching order of events, first behaviour events then the ones of cartService.

PR #193, opened 2020-02-17
search: extend suggestions with type and additional attributes

tlayh:
Adding CategorySuggestions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,5 +18,3 @@
 - category:
     - Tree object uses a Tree Entity now which contains NOT all category properties. You have to fetch the category details separate on demand:
         - search for usages of the data funcs - they may need changes in rendering the data: `data('category´´..`
-- search:
-    - Search module domain model has the new Entity CategorySuggestion that contains suggestions used in the live search and are part of the Result struct.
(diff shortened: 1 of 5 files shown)


tlayh:
> 1. Please add a Changelog.md to the search module (for an example have a look at the cart module) mentioning the new struct field.
> 2. Please add the new field also to the `SearchResult` struct of the `application/searchService` and map it correctly in the `Find` and `FindBy` methods

Added the required changes.

danielpoe:
Looks a bit specific - not sure if this is worth adding new types to commerce core?
Can Suggestion be used in the way how Documents are used (specific types are up to the using module)

tlayh:
> Looks a bit specific - not sure if this is worth adding new types to commerce core?
> Can Suggestion be used in the way how Documents are used (specific types are up to the using module)

I think other projects could benefit from category suggestions as well. Do you have any good idea how to do this different on a project specific level?

carstendietrich:
@tlayh as talked:

please remove the new struct and add a `Type string` and `AdditionalAttributes map[string]string` to the existing `Suggestion` struct. Add also consts for Suggestion Type Product and Suggestion Type Category. Maybe we also add brands right away.

PR #197, opened 2020-03-05
datalayer: fix panic if not able to build absolute url

SchiffFlieger:
If the router fails to build an absolute url, the datalayer would cause a panic. To avoid that, we now check the error and provide an empty url instead.

### DIFF
--- a/w3cdatalayer/application/factory.go
+++ b/w3cdatalayer/application/factory.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha512"
 	"encoding/base64"
 	"flamingo.me/flamingo-commerce/v3/cart/domain/decorator"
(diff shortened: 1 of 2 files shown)


w3cdatalayer/application/factory.go
```diff
@@ -54,6 +59,7 @@ func (s *Factory) Inject(
 	},
 ) {
 	s.router = router2
+	s.logger = logger
```
carstendietrich: Could you please add details about to module to the logger?
`logger.WithField(flamingo.LogKeyModule, "w3cdatalayer")`

SchiffFlieger: Good point. Added it like you said.

PR #202, opened 2020-03-23
checkout: Add the possibility of external source locations ids

carstendietrich:
This enhances the SourcingServiceDetail with an optional external source location id. This is needed if an external system consumes the source locations and needs special ones.

Further more SetSourcesForCartItems function now uses the new SourcingServiceDetail functionality if possible.

TODO:
- [x] update changelog


cart/domain/cart/cartServicePorts.go
```diff
@@ -87,9 +87,13 @@ type (
 
 	// ItemUpdateCommand defines the update item command
 	ItemUpdateCommand struct {
-		// Source Id of where the items should be initial picked - This is set by the SourcingLogic
-		SourceID       *string
-		Qty            *int
+		// SourceID of where the items should be initial picked - This is set by the SourcingLogic
+		SourceID *string
+		// ExternalSourceID contains the picking location used by an external system
+		ExternalSourceID *string
```
danielpoe: SourceId stays the unique key for a Source - right?
So it feels unlogical to add an additional property here since ExternalSourceId will always be determined by the SourceId.

danielpoe: We even discussed to move sourcing code completely from cart package - it can be done by the CartAdapters
(PlaceOrderAdapter OMS is doing sourcing himself / Magento Adapter can do it also)

carstendietrich: > SourceId stays the unique key for a Source - right?
> So it feels unlogical to add an additional property here since ExternalSourceId will always be determined by the SourceId.

Good point, I guess we'll then use a field in the `AdditionalData` property.

carstendietrich: > We even discussed to move sourcing code completely from cart package - it can be done by the CartAdapters
> (PlaceOrderAdapter OMS is doing sourcing himself / Magento Adapter can do it also)

yeah, maybe something we can do for v4 of flamingo-commerce, i'll add an issue.

PR #208, opened 2020-04-02
product: Expose codeLabel via GraphQL and add a method to get attributes by keys

lifedraft:
* Exposing codeLabel property to product attribute type in GraphQL
* Added AttributesByCode method to attributes type to filter attributes by code
* Exposed getAttributesByCode method in GraphQL


product/domain/productBasics.go
```diff
@@ -601,3 +601,15 @@ func (a Attributes) Attribute(key string) Attribute {
 	attribute := a[key]
 	return attribute
 }
+
+// AttributesByCode returns slice of attributes by given attribute codes
+func (a Attributes) AttributesByCode(codes []string) []Attribute {
```
carstendietrich, line 606: I guess the Attributes are key-value pairs. I would therefore recommend to call the method `AttributesByKeys`

carstendietrich: Otherwise please change the body of the method and actually compare the attributes code..

PR #209, opened 2020-04-03
rearrange CHANGELOG.md so most recent version is at top

carstendietrich:
Most recent changes / versions should be at the top of the CHANGELOG, this reorders the changelog entries.


carstendietrich:
@lifedraft as you suggested I rearranged the changelog.

PR #211, opened 2020-04-07
payment: introduce constant for duplicate idempotency key error

zohmi:


payment/domain/payment.go
```diff
@@ -54,16 +54,18 @@ type (
 )
 
 const (
-	// PaymentErrorCodeFailed error
+	// PaymentErrorCodeFailed error will be returned when a general error occured
```
carstendietrich, line 57: please fix typo

PR #215, opened 2020-04-22
checkout: optional cart validation before place order

tessig:
When using the place order state machine, one of the first steps during the process is already validating the cart.
Afterwards the cart is closed and stored inside the context store.
When then later placing the order the OrderService performs the cart validation again.

At this time the cart is already closed and if a validator tries to fetch the cart from the session, an empty cart returns which can lead to invalid validation results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
   * **Breaking**: renamed the following GraphQL types
     * type `Commerce_Cart_BillingAddressFormData` is now `Commerce_Cart_AddressForm`
     * input `Commerce_BillingAddressFormInput` is now `Commerce_Cart_AddressFormInput`
+ 
+**checkout**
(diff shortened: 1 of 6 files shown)


danielpoe:
Root cause seems that a validator access the cart in the "session" and not in the one saved in the context right?

I guess the rules that need to "see" the cart need to be improved or did I get it wrong?

tessig:
@danielpoe That's right. But since flamingo-commerce has no control over the implementations of the validator port, this is a step to keep compatibility.

danielpoe:
Hey understood. But the Validator Port already passed the decorated Cart so the implementation should use this instead using a side effect. Maybe we create a follow up on the Adapter side?

Ist the cart that is passed here already the cart in the placeorder state context?

tessig:
Yes, the passed cart comes from the placeorder state context when using the place order state machine

PR #217, opened 2020-04-24
product: Add LoyaltyEarning property to product model

carstendietrich:
Currently the product domain model only supports paying with loyalty programs, it would be nice to also support a potential earning of different loyalty points/currencies.

This adds LoyaltyEarning properties to the product model which then can later be used to display on the product detail page or in the list view.

- [x] Update changelog
- [x] Add new fields in GraphQL Schema
- [x] Update product model UML


danielpoe:
Where do the LoyaltyEarnings come from? I ask myself if that should go in the product context or if that can be something separate? 
Also just having a Rate is not sufficient for loyalty calculation.

PR #219, opened 2020-04-28
checkout: Add metrics to the standard place order process.

C4rter:
checkout: Add metrics to the standard place order process.

The following open census metrics are now exposed:
-  no of starts w/ new uuid
    - flamingo_standard_place_order_new_uuid{area=""} 1
-  no of runs for each state (bucket)
    - example for a state: flamingo_standard_place_order_processed_state{area="-",state="CompleteCart"} 3
-  no of failed state transitions by state name (bucket)
    - example for a state: flamingo_standard_place_order_failed_state_transition{area="-",state="CompleteCart"} 1

### DIFF
--- a/checkout/application/placeorder/coordinator.go
+++ b/checkout/application/placeorder/coordinator.go
@@ -65,12 +65,12 @@ var (
 	maxLockDuration = 2 * time.Minute
 
 	// newUUIDCount counts generation of new UUIDs
-	newUUIDCount = stats.Int64("flamingo/standard_place_order/new_uuid", "Count of logs", stats.UnitDimensionless)
+	newUUIDCount = stats.Int64("flamingo-commerce/checkout/placeorder/starts", "", stats.UnitDimensionless)
 )
 
 func init() {
 	gob.Register(process.Context{})
-	err := opencensus.View("flamingo/standard_place_order/new_uuid", newUUIDCount, view.Count())
+	err := opencensus.View("flamingo-commerce/checkout/placeorder/starts", newUUIDCount, view.Count())
 	if err != nil {
 		panic(err)
 	}
--- a/checkout/application/placeorder/coordinator.go
+++ b/checkout/application/placeorder/coordinator.go
@@ -66,13 +66,13 @@ var (
 
 	maxLockDuration = 2 * time.Minute
 
-	// newUUIDCount counts generation of new UUIDs
-	newUUIDCount = stats.Int64("flamingo-commerce/checkout/placeorder/starts", "", stats.UnitDimensionless)
+	// startCount counts starts of new place order processes
+	startCount = stats.Int64("flamingo-commerce/checkout/placeorder/starts", "Counts how often a new place order process was started", stats.UnitDimensionless)
 )
 
 func init() {
 	gob.Register(process.Context{})
-	err := opencensus.View("flamingo-commerce/checkout/placeorder/starts", newUUIDCount, view.Count())
+	err := opencensus.View("flamingo-commerce/checkout/placeorder/starts", startCount, view.Count())
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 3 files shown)


checkout/application/placeorder/coordinator.go
```diff
@@ -60,10 +63,17 @@ var (
 	ErrAnotherPlaceOrderProcessRunning = errors.New("ErrAnotherPlaceOrderProcessRunning")
 
 	maxLockDuration = 2 * time.Minute
+
+	// newUUIDCount counts generation of new UUIDs
+	newUUIDCount = stats.Int64("flamingo/standard_place_order/new_uuid", "Count of logs", stats.UnitDimensionless)
```
carstendietrich, line 68: To match the other exposed metrics I would suggest changing the namespace to something like:
`flamingo-commerce/checkout/placeorder/`

In addition this metric isn't really speaking, yes there will be a new uuid but that doesn't express what that means. In this case a new place order process is started. So i would rename from `new_uuid` to `new` or `starts`. What is the second parameter used for `"Count of logs"`?

checkout/domain/placeorder/process/process.go
```diff
@@ -70,12 +76,27 @@ type (
 	}
 )
 
+var (
+	// processedState counts processed states
+	processedState = stats.Int64("flamingo/standard_place_order/processed_state", "Count of processed states", stats.UnitDimensionless)
```
carstendietrich, line 81: would change the namespace here as well

checkout/domain/placeorder/process/process.go
```diff
@@ -166,6 +187,9 @@ func (p *Process) Run(ctx context.Context) {
 		return
 	}
 
+	censusCtx, _ := tag.New(ctx, tag.Upsert(opencensus.KeyArea, "-"), tag.Upsert(keyState, currentState.Name()))
```
carstendietrich, line 190: Why overriding the area here? I guess we could just remove that.

checkout/domain/placeorder/process/process.go
carstendietrich: maybe rename to state_run_count

checkout/domain/placeorder/process/process.go
```diff
@@ -70,12 +76,27 @@ type (
 	}
 )
 
+var (
+	// processedState counts processed states
+	processedState = stats.Int64("flamingo/standard_place_order/processed_state", "Count of processed states", stats.UnitDimensionless)
+	// failedStateTransition counts failed state transitions
+	failedStateTransition = stats.Int64("flamingo/standard_place_order/failed_state_transition", "Count of processed states", stats.UnitDimensionless)
```
carstendietrich, line 83: maybe rename to state_failed_count

checkout/domain/placeorder/process/process.go
C4rter: Wasn't sure what to do with this "area". If it's fine that it's empty, that's okay for me as well.

carstendietrich: jeah just remove the whole `tag.Upsert(opencensus.KeyArea, "-"),` since it isn't any value.

checkout/application/placeorder/coordinator.go
C4rter: Removed the description now on all three metrics.

checkout/domain/placeorder/process/process.go
C4rter: Injected the area config now as discussed and added the area to the metrics.

checkout/domain/placeorder/process/process.go
C4rter: Renamed that.

checkout/domain/placeorder/process/process.go
C4rter: Renamed that.

PR #222, opened 2020-05-19
product: Add API controller

danielpoe:
Products Module should get a Rest API endpoint to get products

### DIFF
--- a/product/module.go
+++ b/product/module.go
@@ -3,8 +3,6 @@ package product
 import (
 	"flamingo.me/dingo"
 	"flamingo.me/flamingo-commerce/v3/price"
-
-	//"flamingo.me/flamingo-commerce/v3/price"
 	"flamingo.me/flamingo-commerce/v3/product/domain"
 	"flamingo.me/flamingo-commerce/v3/product/infrastructure/fake"
 	"flamingo.me/flamingo-commerce/v3/product/interfaces/controller"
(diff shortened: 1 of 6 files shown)


product/module.go
```diff
@@ -80,3 +87,16 @@ func (r *routes) Routes(registry *web.RouterRegistry) {
...
+type apiroutes struct {
+	apiController *controller.ApiController
+}
+
+func (r *apiroutes) Inject(apiController *controller.ApiController) {
+	r.apiController = apiController
+}
+
+func (r *apiroutes) Routes(registry *web.RouterRegistry) {
+	registry.Route("/api/v1/products/:marketplacecode", "products.api.get")
```
carstendietrich, line 100: In general I am a fan of API versioning, but the existing API routes do not have it at the moment. So I'm not sure if we should remove it for the sake of consistency.

product/module.go
```diff
@@ -69,7 +76,7 @@ type routes struct {
 	controller *controller.View
 }
 
-func (r *routes) Inject(controller *controller.View) {
+func (r *routes) Inject(controller *controller.View, apiController *controller.ApiController) {
```
carstendietrich, line 79: please remove unused param

product/module.go
```diff
@@ -80,3 +87,16 @@ func (r *routes) Routes(registry *web.RouterRegistry) {
 	h, _ = registry.Route("/product/:marketplacecode/:variantcode/:name.html", `product.view(marketplacecode, variantcode, name, backurl?="")`)
 	h.Normalize("name")
 }
+
+type apiroutes struct {
```
carstendietrich, line 91: Please use proper camel case here.

product/module.go
carstendietrich: Also why use /product**s**/ and not /product/  like the normal routes?

product/module.go
```diff
@@ -60,6 +64,9 @@ commerce: {
 			enabled: bool | *false
 			currency: *"€" | !=""  
 		}
+		api: {
+			enabled: bool | *true
```
carstendietrich, line 68: Whats the point of this config flag? Flamingo commerce already has a ton of configs, in my opinion we should try to reduce them instead of adding more.

product/module.go
danielpoe: I wanted to add "v1" as additional routes for the existing apis.

Regarding plural: the rest api styles I know prefer plural for collections.
The only other api thats in flamingo-commerce is "cart" and there is no plural since its not a collection but more a "current cart" thing.
What do you think?

product/module.go
danielpoe: The possibility to switch of this endpoint.
I thought that it might expose undesired data (like all possible prices)

product/module.go
carstendietrich: ah okay had to take a look at REST best practices my self. Plural is fine 👍 
Also saw your other pull requests containing the v1 prefix. thats fine for me as well.

product/module.go
carstendietrich: Okay, guess we have the same "problem" within graphql.

PR #223, opened 2020-05-20
Add swagger spec

danielpoe:
The REST APIs exposed by Flamingo Commerce should get Doc comments that allow tools like swag to parse them.

Also we should add API versioning


bastianccm:
I would like to autogenerate more, e.g. the "consume/produce" values are not always correct, and the whole concept (consume/produce) is currently not sufficiently implemented.

Also directly integrating it as a web.RouterModule might be nice? We would hit "2 flies with one hit", eliminate duplicate code/documentation, and thus lower the bar and reduce possible future bugs

commerce.go
```diff
@@ -0,0 +1,10 @@
+package main
```
bastianccm, line 1: from my experience i would highly recommend to not put this here as a main package, as it will throw errors as there is no `main` method. `go get flamingo.me/flamingo-commerce/v3` would fail because it would try to compile a non-existing binary

cart/interfaces/controller/cartapicontroller.go
```diff
@@ -74,6 +74,13 @@ func (cc *CartAPIController) Inject(
 }
 
 // GetAction Get JSON Format of API
+// @Summary Get the current cart
+// @Tags cart
+// @Accept json
```
bastianccm, line 79: This does not accept json, right? It's a GET request

commerce.go
```diff
@@ -0,0 +1,10 @@
...
+
+//go:generate go run github.com/swaggo/swag/cmd/swag init --generalInfo=commerce.go --output=docs/openapi
```
bastianccm, line 3: This won't nicely integrate with customizations.
I think it might make more sense to integrate the swag-concept with a generic API module.
The API module can be found by swag (using dingo) and implement a route-module, to auto-generate as much as possible

danielpoe:
> I would like to autogenerate more, e.g. the "consume/produce" values are not always correct, and the whole concept (consume/produce) is currently not sufficiently implemented.
> 
> Also directly integrating it as a web.RouterModule might be nice? We would hit "2 flies with one hit", eliminate duplicate code/documentation, and thus lower the bar and reduce possible future bugs

We talked about this. "swag" seems to be one of the only active generators working with go modules.
The duplicate informations is mainly path/http method - most of the other infos (return status and types) need to be annotated somewhere anyhow.

Result for now was: 
* Use swag as is and accept duplicate infos (like path)
* Make clear in the docs that this are swago annotations used to generate swagger
* Later: think of autogenerating an apirouter type from the comments
* Later: autotest the generated swagger against running api to make sure spec matches implementation

commerce.go
danielpoe: Moved to docs/openapidoc.go

danielpoe: Moving to subfolder did not work (generation failed)... I added it as main_test..

Other suggestions @bastianccm

openapidoc.go
```diff
@@ -0,0 +1,15 @@
+package main_test
```
bastianccm, line 1: Rename to `package commerce` (as discussed)

danielpoe: Done - pleas recheck

PR #231, opened 2020-06-03
Add configurable support to graphQL / swaggerdoc improvements

danielpoe:


danielpoe:
> Looks good, is the swago output now stable?
> If so we could get rid of the Travis CI workaround: https://github.com/i-love-flamingo/flamingo-commerce/blob/master/.travis.yml#L29

Kind of the issue is still open but the main things are fixed.. I would fix this in a seperate PR (linked to the issue) later

carstendietrich:
> > Looks good, is the swago output now stable?
> > If so we could get rid of the Travis CI workaround: https://github.com/i-love-flamingo/flamingo-commerce/blob/master/.travis.yml#L29
> 
> Kind of the issue is still open but the main things are fixed.. I would fix this in a seperate PR (linked to the issue) later

sounds good, will merge for now.

Issue #234, opened 2020-06-08
Different formatting of attribute values returned by Value() and Values() of Attribute of BasicProductData


richard-ra:
Hi. 
While investigating a bug which was caused by the fact that Value() returns the raw value of the attribute but with whitespaces trimmed on both ends, I noticed that the Values() method returns a list of attribute values but the values do not have the whitespaces removed. 

https://github.com/i-love-flamingo/flamingo-commerce/blob/bf76717aebb7d4ea1e81bdd8acbe70a307b2f520/product/domain/productBasics.go#L225-L227

and

https://github.com/i-love-flamingo/flamingo-commerce/blob/bf76717aebb7d4ea1e81bdd8acbe70a307b2f520/product/domain/productBasics.go#L264-L273

danielpoe:
Hey 
Thanks that is inconsitent.
See attached PR for a fix.

You had problems with the Trimming in general?

(We are planning to adjust the attribute modelling in the future simelar to the attributs model in caegory - but since this is breaking thats up for v4)

richard-ra:
Hi,

The only reason I realised that Value() returns the value with trimmed whitespaces was because we have the case where we compare the return value of Value() to the attribute value/s in 'variantVariationAttributesSorting':
https://github.com/i-love-flamingo/flamingo-commerce/blob/c440e4ff91ef5db296e01ea0d994e07b3112777e/product/domain/productTypeConfigurables.go#L32

Since the source of the attribute data is the same, this means that when the comparison  happened, we never got a match if the attribute value had a ' ' at the beginning or end. The solution in this case was to also trim whitespaces for the attribute values in 'variantVariationAttributesSorting' before comparing.

Furthermore, this issue only occurred because of a user error when managing the attribute value in the PIM where a whitespace was entered at the beginning of the value by mistake.

Thank you and let me know if there are any more questions

danielpoe:
Hey thank you for the Details.
So the attached branch should fix that right?

danielpoe:
Any other support needed please let us know

richard-ra:
I looked at it and it looks fine but isn't the newly added 'else' statement a breaking change as currently "Values()" returns an empty array if rawValue is a single value and that statement would make the array always contain a value? I'm currently doing frontend development (at AOE) so go is not my strong suit so I might be wrong.

PR #236, opened 2020-06-11
product: #234 make values return trimmed as well

danielpoe:


product/domain/productBasics_test.go
```diff
@@ -80,7 +80,8 @@ func TestAttributeValues(t *testing.T) {
 	a := Attribute{RawValue: "some string"}
 	result := a.Values()
 	assert.IsType(t, []string{}, result)
-	assert.Empty(t, result)
+	assert.Len(t, result, 1)
+	assert.Equal(t, "some string", result[0])
```
tessig, line 84: please add this kind of assertion also for the slice case

carstendietrich: updated the test please have a look

tessig (approved):

PR #240, opened 2020-06-17
checkout: Improve controller error handling

carstendietrich:
In case of error during payment always redirect back to checkout/review page in stead of rendering the template under a different route

Currently in case of a error during payment/placing the order we just render the checkout / review template again on the same route.

If the user then reloads the page he is still on the placeorder route and will trigger the error again or cause other potential side effects.
Its therefore better to send him back on the checkout route and render the error in the matching action. So that the customer then can start a new try.

### DIFF
--- a/checkout/interfaces/controller/checkoutcontroller.go
+++ b/checkout/interfaces/controller/checkoutcontroller.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/gob"
 	"errors"
-	"fmt"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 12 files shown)


checkout/interfaces/controller/checkoutcontroller.go
```diff
@@ -546,13 +549,22 @@ func (cc *CheckoutController) ReviewAction(ctx context.Context, r *web.Request)
 		DecoratedCart: *decoratedCart,
 	}
 
+	flashErrors := r.Session().Flashes(CheckoutErrorFlashKey)
```
thehl, line 552: You use nearly the same code block in line 380ff except the return, does it make sense to build a private function to avoid code duplication?

carstendietrich: Moved logic to a helper func.

PR #245, opened 2020-07-03
Extend GraphQL customer information with login status.

lifedraft:
Unfortunately by calling the Commerce_Customer query this failed for not authenticated users. From Frontend perspective this query should be more resilient, should not fail and return users login state. Therefore this MR extends the GraphQL response of this query.

TBD; We could remove Commerce_Customer_Status query in future major release as this becomes obsolete?


customer/interfaces/graphql/schema.graphql
```diff
@@ -4,7 +4,9 @@ type Commerce_Customer_Status_Result {
 }
 
 type Commerce_Customer_Result {
-    id: String!
+    id: String! @deprecated(reason: "Use `userID` field.")
+    userID: String!
+    isLoggedIn: Boolean!
```
bastianccm, line 9: isLoggedIn is semantically not making sense here

customer/interfaces/graphql/schema.graphql
```diff
@@ -4,7 +4,9 @@ type Commerce_Customer_Status_Result {
 }
 
 type Commerce_Customer_Result {
-    id: String!
+    id: String! @deprecated(reason: "Use `userID` field.")
+    userID: String!
```
bastianccm, line 8: What is the difference between userID and id here? Why is it deprecated? "Use another field" is not a reason

customer/interfaces/graphql/queryresolver.go
```diff
@@ -46,12 +47,14 @@ func (r *CustomerResolver) CommerceCustomerStatus(ctx context.Context) (*dtocust
 // CommerceCustomer resolver the commerce customer
 func (r *CustomerResolver) CommerceCustomer(ctx context.Context) (*dtocustomer.CustomerResult, error) {
 	user, err := r.service.GetForIdentity(ctx, web.RequestFromContext(ctx))
-	if err != nil {
-		return nil, err
+	if errors.Is(err, application.ErrNoIdentity) {
+		return &dtocustomer.CustomerResult{IsLoggedIn: false}, nil
```
bastianccm, line 51: Simply do `return nil, nil`, to indicate that there is no Customer, but also no error.

customer/interfaces/graphql/dtocustomer/dto.go
```diff
@@ -14,6 +14,8 @@ type (
 	// CustomerResult is a dto
 	CustomerResult struct {
 		ID                     string
+		UserID                 string
+		IsLoggedIn             bool
```
bastianccm, line 18: See the GraphQL Schema for comments

customer/interfaces/graphql/schema.graphql
lifedraft: removed.

customer/interfaces/graphql/queryresolver.go
lifedraft: changed.

customer/interfaces/graphql/dtocustomer/dto.go
lifedraft: also removed.

PR #248, opened 2020-07-15
Add configuration option for multistore setup to be able to have to different cart caches

tlayh:
Currently, it is possible to have multiple carts. This pull requests adds a configuration option to have multiple cart session caches by adding a prefix. This should be non-breaking, because the configuration is optional and uses the constants if no option is set.


tlayh:
Found a much easier solution

danielpoe:
> Found a much easier solution

Good to hear - how?

Different Domains to be sure sessions do not interfer?

PR #249, opened 2020-07-16
product: update fake service to contain loyalty pricing

carstendietrich:


product/infrastructure/fake/productservice.go
```diff
@@ -117,16 +115,19 @@ func (ps *ProductService) Get(ctx context.Context, marketplaceCode string) (doma
 		return product, nil
 	}
 	if marketplaceCode == "fake_simple" {
-		return ps.FakeSimple(marketplaceCode, false, false, false), nil
+		return ps.FakeSimple(marketplaceCode, false, false, false, true, false), nil
```
tessig, line 118: we could finally use a switch here

carstendietrich: indeed will update

PR #262, opened 2020-09-23
product: Fallback variants in variationSelection should respect the sorting

tomraithel:
The new `variationSelection` provides a fallback variant for each option, if no *exact* match could be found with the given attributes.

*Current behaviour*
If no matching variant has been found, the fallback variant is the first best match from the *variants*  array.

*Expected behaviour*
The fallback variant should respect the sorting provided in `variationAttributesSorting`


tomraithel:
Removed `VariationSelectionOption.MarketPlaceCode` with `VariationSelectionOption.Variant` to be able to extend it

product/interfaces/graphql/product/dto/variantsToVariationSelectionsMapper.go
```diff
@@ -134,53 +167,48 @@ func (m *variantsToVariationSelectionsMapper) hasVariantsWithMatchingAttributes(
...
+
+	attributeGroup.eachAttributeInOrder(func(attribute *domain.Attribute) {
+		var state VariationSelectionOptionState
+		var matchingVariant domain.Variant
+
+		if m.hasActiveVariant() {
+			mergedAttributes := m.getActiveAttributesWithOverwrite(*attribute)
+			exactMatchingVariant := m.findMatchingVariant(mergedAttributes)
+
+			if exactMatchingVariant != nil {
```
carstendietrich, line 179: Is all the if/else branching necessary? looks like a lot of cognitive load to me. is it possible to decrease that using early returns/always doing the else part?

tomraithel: Yeah I know, i´ts the most ugly method in this whole class because the logic behind it is inherently complex. I could at least try to split it up in multiple methods to make it a bit more readable

tomraithel: I got rid of some ifs by separating the logic into smaller methods. I think now it´s better to grasp what is getting on.

carstendietrich: Jeah way better! Added a small improvement, will merge when pipeline is green. 👍

carstendietrich (approved):

product/interfaces/graphql/product/dto/variantsToVariationSelectionsMapper.go
```diff
@@ -169,12 +169,10 @@ func (m *variantsToVariationSelectionsMapper) buildVariationSelectionOptions(att
 	var options []VariationSelectionOption
 
 	attributeGroup.eachAttributeInOrder(func(attribute *domain.Attribute) {
-		var option *VariationSelectionOption
+		option := m.createOptionWithoutActiveVariant(*attribute)
```
tomraithel, line 172: @carstendietrich hmm, by removing this else branch you will always loop through all options twice when an active variant is set. The result is the same, but this is probably not what we want in terms of performance.

carstendietrich: I am aware of this, I guess it is more important to have the code easily understood in a few months than to have it micro-optimized. If there was an API call or a large calculation, I would stick to the else branch as well.

tomraithel: Ok fine then

PR #263, opened 2020-10-01
product: Add feature to the product fakeservice to use json files as the source.

lifedraft:
Currently, fake services are very static, in case you need something special, a custom attribute, variant options, etc. developers have to either create a pull request or implement own fake service in their projects. This feature allows developers to define/extend current fake products with their own versions of it.

### DIFF
--- a/product/infrastructure/fake/productservice.go
+++ b/product/infrastructure/fake/productservice.go
@@ -374,15 +374,10 @@ func (ps *ProductService) getProductFromJSON(code string) (domain.BasicProduct, 
 
 // jsonProductCodes returns an ordered list of the json product codes
 func (ps *ProductService) jsonProductCodes() []string {
-	productCodes := make([]string, 0, len(ps.testDataFiles))
 	keys := make([]string, 0, len(ps.testDataFiles))
 	for k := range ps.testDataFiles {
 		keys = append(keys, k)
 	}
 	sort.Strings(keys)
-	for _, k := range keys {
-		productCodes = append(productCodes, k)
-	}
-
-	return productCodes
+	return keys
 }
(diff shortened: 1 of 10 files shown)


product/module.go
```diff
@@ -75,7 +75,8 @@ commerce: {
 		slugAttributeCode: string | *"urlSlug"
 		fakeservice: {
 			enabled: bool | *false
-			currency: *"€" | !=""  
+			currency: *"€" | !=""
+			jsonTestDataFolder: *"testdata/products" | !=""
```
tessig, line 79: Here, I must specify a `jsonTestDataFolder` even if the feature is disabled.
I suggest to use an if statement here or make it an optional value, which would match the injection.

tessig (changes requested):
Good work so far.
Please also add a couple of sentences to the Readme of the product module to document the new feature.
Please also add a test at least for the json loading part, e.g. by testing the `Get` with activated fake service and a small json file.

product/infrastructure/fake/productservice.go
```diff
@@ -366,12 +372,16 @@ func (ps *ProductService) getProductFromJSON(code string) (domain.BasicProduct,
...
 	productCodes := make([]string, 0, len(ps.testDataFiles))
-
-	for key := range ps.testDataFiles {
-		productCodes = append(productCodes, key)
+	keys := make([]string, 0, len(ps.testDataFiles))
+	for k := range ps.testDataFiles {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, k := range keys {
```
tessig, line 383: why not directly return `keys`?

product/module.go
```diff
@@ -76,7 +76,7 @@ commerce: {
 		fakeservice: {
 			enabled: bool | *false
 			currency: *"€" | !=""
-			jsonTestDataFolder: *"testdata/products" | !=""
+			jsonTestDataFolder?: !=""
```
tessig, line 79: now we loose the default value. I'm not sure either what would be the most comfortable way, but how about something like this
```
if enabled {
  jsonTestDataFolder: string | *"testdata/products"
}
```

tessig (approved):

PR #264, opened 2020-10-07
category: Add JSON based fake category service

kalypso-byte:
Provide a fake service for mocking categories and the category tree with json files as well as a standard set that can be used as is.


category/application/fake/category_service.go
```diff
@@ -0,0 +1,75 @@
+package fake
```
carstendietrich, line 1: Usually our fake implementations life under infrastructure/fake/ we should move this as well to be more inline with the other modules.

category/application/fake/category_service.go
```diff
@@ -0,0 +1,75 @@
...
+	logger        flamingo.Logger
+}
+
+var _ domain.CategoryService = new(CategoryService)
+
+// Inject dependencies
+func (f *CategoryService) Inject(logger flamingo.Logger, config *struct {
+	TestDataFolder string `inject:"config:commerce.category.fakeService.testDataFolder,optional"`
+}) {
+	f.logger = logger.WithField(flamingo.LogKeyModule, "category")
```
carstendietrich, line 25: Maybe provide the `flamingo.LogKeyCategory` with something like `fakeService` so that filtering logs later is easier.

category/application/fake/category_service.go
```diff
@@ -0,0 +1,75 @@
...
+	index := findTreeRootIndex(activeCategoryCode, categoryTree)
+	if index == -1 {
+		return nil, domain.ErrNotFound
+	}
+	return categoryTree[index], nil
+}
+
+// Get the category of the given category code
+func (f CategoryService) Get(_ context.Context, categoryCode string) (domain.Category, error) {
+	category := LoadCategory(categoryCode, f.testDataFiles, f.logger)
```
carstendietrich, line 45: Currently feels like an overkill to need separate files for each category since that information is already in the categroyTree.json.. Do you plan on providing more detailed information in the single JSON? otherwise, we could just traverse through the category tree and pick the matching category with all its information from there. Maybe easier for testers to edit just one file.

carstendietrich (changes requested):
Nice addition, left some feedback. Please also update the CHANGELOG.md in the project root and mention the new fake service in the readme of the category module.

If you have some spare time it would be awesome if you could switch the category module to the new Cue config style and add all possible configurations so it is clearer what configs are supported.

carstendietrich (approved):

PR #269, opened 2020-10-26
checkout: Add support for wallet payments

carstendietrich:
Currently flamingo-commerce only supports confentional payment methods like hosted payment pages or iframes, a nice addition would be the wallet payment provided by Apple or Google. Based on the open PaymentRequest API specifications a new state was added.


masmrlar (approved):
Approved - nice work

PR #270, opened 2020-10-28
product: Enhance fake search service

kalypso-byte:


product/infrastructure/fake/searchservice.go
```diff
@@ -23,29 +23,81 @@ func (s *SearchService) Inject(
...
+
+	// - try finding product by marketPlaceCode given in query
+	if query, found := s.filterValue(filters, "q"); found {
+		var product domain.BasicProduct
+		if len(query) > 0 {
+			product, _ = s.productService.Get(ctx, query[0])
+		}
+		if product != nil {
+			products = append(products, product)
+		}
```
carstendietrich, line 72: Why not return here immediately? With this implementation, we currently can't test queries that don't match any product which could be a valid case you want to simulate with a fake service.

tessig (approved):

carstendietrich (approved):

PR #271, opened 2020-10-28
introduce github actions

tessig:
start with the simple test suite


carstendietrich (approved):
Cool, let's take this as a starting point and move from Travis to GitHub Actions.

PR #280, opened 2020-11-10
sourcing: Add deliveryInfo to interface StockProvider GetStock

masmrlar:
Implementing StockProviders require the deliveryInfo for specific stock checks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,11 @@
     * The product fake search service returns no products if it is queried with `no-results`
 * Expose `VariantVariationAttributesSorting` on `domain.ConfigurableProductWithActiveVariant`
 
+**sourcing**
+* **Breaking**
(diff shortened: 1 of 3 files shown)


danielpoe:
Can you give details for the use case?

masmrlar:
@danielpoe a usecase for this is for example a requirement where the StockProvider has to use a specific logic based on product data/attributes and the deliveryInfo (like DeliveryCode) - to override stock data. 

And a pretty specific usecase is: we want to have a delivery specific alwaysInStock logic - let's call it `alwaysInStockForDelivery` beneath `alwaysInStock`.

carstendietrich (changes requested):
I guess it can't hurt that the stock also gets the delivery info, never the less it's a breaking change. Please, update the project's main CHANGELOG.md and mark this change as breaking.

carstendietrich (approved):
Looks fine.

PR #281, opened 2020-11-18
cart: Trying to fix race on in memory cart storage

masmrlar:
The init func of InMemoryCartStorage is intended to init the storage if required as well as the sync.Locker to prevent concurrent operations on the storage. 

This handling does not solve the concurrent calls, as "it is already too late", a locking must be done before the first access to the storage map is done. 

The refactored implementation uses Inject to ensure that storage and Locker have been initialized to correctly lock on each receiver.


carstendietrich (approved):
Nice one!

PR #287, opened 2020-11-26
customer graphql: Add field for getting specific address from customer

kalypso-byte:


carstendietrich (changes requested):
@kalypso-byte please regenerate the GraphQL files of the integration test, see makefile.

carstendietrich (approved):

PR #290, opened 2020-12-01
checkout: resolve goroutine leak in redis locker

tessig:
`ticker.C` does not close on stop, so all routines with that range were running forever


carstendietrich (approved):
Looks good, didn't know about goleak 👍🏻

PR #292, opened 2020-12-03
product: Add support for product badges

masmrlar:
Badges on product data can be used in frontend to highlight products with additional informations.

As an example a badge can be used to mark a product as `new` or `special price` product.

Badges are provided on BaseData and on TeaserData of a product.

Special handling:
- configurable products badges are the ones set on the configurable "parent" product. they are not collected over all children of a configurable.
- configurables with active variant badges are the ones set on the active variant (not on the configurable parent)


carstendietrich (approved):
Looks good, added the missing changelog entry. 👌🏻

PR #293, opened 2020-12-09
product: Add convenience function to get first product badge

masmrlar:
Graphql requires to get the first product badge if available. A convenience function is added to the badges as well as to the graphql schema.


product/domain/productBasics.go
```diff
@@ -665,3 +668,17 @@ func (c *CategoryTeaser) CPath() string {
 	}
 	return c.Parent.CPath() + "/" + c.Code
 }
+
+// GetFirst of the badges, returns nil if there is no first badge
+func (b Badges) GetFirst() *Badge {
+	if nil == b {
```
carstendietrich, line 674: Can be removed since `len(nil) = 0`

product/domain/productBasics.go
```diff
@@ -665,3 +668,17 @@ func (c *CategoryTeaser) CPath() string {
 	}
 	return c.Parent.CPath() + "/" + c.Code
 }
+
+// GetFirst of the badges, returns nil if there is no first badge
+func (b Badges) GetFirst() *Badge {
```
carstendietrich, line 673:
```suggestion
func (b Badges) First() *Badge {
```

For new stuff, we could adhere to effective Go: https://golang.org/doc/effective_go.html#Getters

product/interfaces/graphql/schema.graphql
```diff
@@ -225,6 +225,7 @@ type Commerce_Product_SearchResult {
 
 type Commerce_Product_Badges {
     all: [Commerce_Product_Badge!]
+    getFirst: Commerce_Product_Badge
```
carstendietrich, line 228:
```suggestion
    first: Commerce_Product_Badge
```
would go with just `first` here as well

carstendietrich (changes requested):
Looks fine to me, only some minor suggestions.

carstendietrich (approved):
🎉

PR #294, opened 2020-12-09
Added `IsAlwaysInStock` to BasicProductData

masmrlar:
The convenience function `IsAlwaysInStock` prevents erratic repetitions in code checking if a product is marked to be always in stock.


carstendietrich:
@masmrlar the always is stock attribute thingy was introduced by accident. we'll remove it completely with an upcoming commerce version. closing here for now.

carstendietrich:
see #295

PR #296, opened 2020-12-23
cart: Retrieve cart when deleting item when updating item qty to update cache

ChristopherSchwab:
In the cartService UpdateItemQty function, in case the new qty is actually smaller than 0, the function calls the DeleteItem function and returns the result. After the DeleteItem function finishes, the UpdateItemQty functions updates the cache in a defered function. The cart used here is not updated with the deletion of the item and even overwrites the cache update of the DeleteItem function.


cart/application/cartService.go
```diff
@@ -271,7 +271,13 @@ func (cs *CartService) UpdateItemQty(ctx context.Context, session *web.Session,
 
 	qtyBefore := item.Qty
 	if qty < 1 {
-		return cs.DeleteItem(ctx, session, itemID, deliveryCode)
+		err := cs.DeleteItem(ctx, session, itemID, deliveryCode)
```
danielpoe, line 274: Look inside the called function - it should update the cache already:
https://github.com/i-love-flamingo/flamingo-commerce/blob/9fde5d432901224600921bfdb55e776032723bd4/cart/application/cartService.go#L408

And the DeleteItem Adapter is supposed to return the new cart: https://github.com/i-love-flamingo/flamingo-commerce/blob/9fde5d432901224600921bfdb55e776032723bd4/cart/application/cartService.go#L426

danielpoe (changes requested):
Imho the PR is not needed

cart/application/cartService.go
ChristopherSchwab: Yes, the DeleteItem function works as intended and the cache is updated afterwards. The problem is that after that the cache is updated again here: https://github.com/i-love-flamingo/flamingo-commerce/blob/9fde5d432901224600921bfdb55e776032723bd4/cart/application/cartService.go#L257

And the cart object used to update the cache does not include the item deletion, effectively overwritting the cache with the cart in a status before the DeleteItem function was executed.

For UpdateItemQty with qty 0:
- Start UpdateItemQty
- Start DeleteItem
- Update cache (DeleteItem) // cache is correct
- Update cache (UpdateItemQty) // cache is incorrect

The PR suggest to fetch the cart again to also have the cache correct after the second update.

jeinfeldt: @danielpoe 
Hola!
Could you revisit the discussion? ChristopherSchwab's point seems valid to me.
On immediate return the cache gets overwritten with the previously assigned `cart` variable
and therefore the cache will not include the current cart.

We need to overwrite the cart variable in order to have the deferred cache update to be valid in this function.

jeinfeldt: I guess it boils down to:
Is the function in the return executed before or after the `defer` is executed?
Will debug locally and in best case provide a test case that fails to present the problem

jeinfeldt: Hey @danielpoe 
I debugged locally and the problem is as ChristopherSchwab described.
See these two code snippets showing the problem:
Current Flow
https://play.golang.org/p/l1g1DJaGtg4

Proposed adjustment
https://play.golang.org/p/SK4yELp9aon

As far as I can understand the behaviour here is the same.

carstendietrich (approved):

cart/application/cartService.go
carstendietrich: Okay, then let's merge this PR ✅

danielpoe: No we make it even more complicated.

What the method actually wanted to do is if the qty=0 then instead of updating the item we deal with the command like Deleting the item.

So why not just move the existing "guard clause" to the start of the method (before the defer) ?

danielpoe (changes requested):

cart/application/cartService.go
carstendietrich: Oh 🙈 

Yeah maybe that the better approach. Will quickly update the code so that we can move on with this PR.

jeinfeldt: True, that would be even cleaner

jeinfeldt: @danielpoe 
Could you check carstendietrich's update?

danielpoe (approved):

PR #298, opened 2021-01-19
product: Add promotions to product search

lebck:


search/domain/service.go
```diff
@@ -102,6 +103,14 @@ type (
 		AdditionalAttributes map[string]string
 	}
 
+	// Promotion result during search
+	Promotion struct {
+		Title   string
+		Content string
+		URL     string
+		// Media []string
```
carstendietrich, line 111: Maybe we use the Media Struct from the product domain model here. And we could also add some `AdditionalAttributes map[string]string` to easily add more meta data if needed

carstendietrich (changes requested):
Integration test seem to fail due to outdated generated code. See the makefile to update the generated code.

carstendietrich (approved):

PR #291, opened 2020-12-03
checkout: always start a new place order process when requested.

tessig:
To hold on the existing process can confuse customers who abandoned there process and want to start a new one but ended up in the abandoned one again.

PR #300, opened 2021-01-29
checkout: Update API controller to match GraphQL resolver

carstendietrich:
In PR #291 we decided to change the behaviour of the StartPlaceOrder mutation to always start a fresh process.

This updates the API controller to match the behavior in the mutation resolver


masmrlar (approved):
Approved, especially the removal of of the implicit refresh and let it be handled explicit.

PR #301, opened 2021-02-08
cart: Add routes for updating/deleting cart items, switch to RESTful naming for v1 cart endpoint

carstendietrich:
Currently, it is not possible to update/delete a single item via the cart API. This PR adds two new actions to the cart API controller.

In addition, the new v1 prefixed cart endpoints are not consistently named when it comes to the routes and used HTTP verbs. I updated those as well to be more RESTful and represent what operations are immutable.


masmrlar (approved):
Nice refactorings.

PR #303, opened 2021-02-15
cart: Update API to actually clean whole cart not only items

carstendietrich:
Currently, the DELETE operation on the cart is misleading, since it's only clearing cart items but not the other information like delivery info or billing address.

I therefore updated the implementation so that the whole cart gets cleaned. The old behaviour is now exposed with a more matching route.

The defaultCartBehaviour was updated to implement the interfaces correctly.


masmrlar (approved):
Nice refactoring / renaming of the api, it's now comprehensive and consistent. Approved.

PR #304, opened 2021-02-16
cart: update comments to match GoDoc style, update plant UML to show more details

carstendietrich:
The cart plant UML was quite outdated, I updated it and added some details.

While checking the Go code in parallel I also fixed some comments.


jeinfeldt (approved):
Beautiful!

PR #307, opened 2021-02-25
checkout: Switch to testcontainers for docker based tests

carstendietrich:
Testcontainers seems to be the advanced alternative to ory/dockertest, the API is similar but testcontainers comes with a reaper that cleans up stale containers even if the test crashes


tessig (approved):

jeinfeldt:
Looks great! As far as I can see https://github.com/i-love-flamingo/flamingo-commerce/issues/305 would still be required though as `github.com/testcontainers/testcontainers-go` also has a dependency to `github.com/opencontainers/runc`
in the vulnerable version

https://github.com/testcontainers/testcontainers-go/blob/master/go.mod#L30

jeinfeldt (approved):

carstendietrich:
> Looks great! As far as I can see #305 would still be required though as `github.com/testcontainers/testcontainers-go` also has a dependency to `github.com/opencontainers/runc`
> in the vulnerable version
> 
> https://github.com/testcontainers/testcontainers-go/blob/master/go.mod#L30

🤦🏻  yeah seems to be right, then lets wait for the fix in core and use your pr to bump the version.

PR #311, opened 2021-04-26
checkout: dispatch events on final place order states

tessig:


checkout/domain/placeorder/states/failed.go
```diff
@@ -4,33 +4,55 @@ import (
...
 	// Failed state
 	Failed struct {
-		Reason process.FailedReason
+		Reason      process.FailedReason
+		eventRouter flamingo.EventRouter
+	}
+
+	// FailedEvent is dispatched when the final failed state runs
+	FailedEvent struct {
+		Reason         process.FailedReason
```
carstendietrich, line 20: Isn't the failed reason already contained in `process.Context`?

checkout/domain/placeorder/states/failed.go
```diff
@@ -4,33 +4,55 @@ import (
...
 	"flamingo.me/flamingo-commerce/v3/checkout/domain/placeorder/process"
+	"flamingo.me/flamingo/v3/framework/flamingo"
 	"go.opencensus.io/trace"
 )
 
 type (
 	// Failed state
 	Failed struct {
-		Reason process.FailedReason
+		Reason      process.FailedReason
```
carstendietrich, line 14: Don't see any usages of this struct field. Maybe we should get rid of it. The failed reason is stored in the overall process context anyways.

carstendietrich (changes requested):

carstendietrich (approved):

PR #312, opened 2021-04-27
cart: introduce custom attributes of cart and additional data of delivery in graphql

el0c1n:


cart/application/cartService.go
```diff
@@ -1200,3 +1200,59 @@ func (qar QtyAdjustmentResults) HasRemovedCouponCodes() bool {
 
 	return false
 }
+
+// UpdateAdditionalData of cart
+func (cs *CartService) UpdateAdditionalData(ctx context.Context, session *web.Session, key string, value string) (*cartDomain.Cart, error) {
```
carstendietrich, line 1205: Would expect here to pass in all additional data and not just a single key, guess the caller could do the merging/updating of a single one.

cart/application/cartService.go
```diff
@@ -1200,3 +1200,59 @@ func (qar QtyAdjustmentResults) HasRemovedCouponCodes() bool {
...
+	cart, defers, err := behaviour.UpdateAdditionalData(ctx, cart, &additionalData)
+	defer func() {
+		cs.updateCartInCacheIfCacheIsEnabled(ctx, session, cart)
+		cs.dispatchAllEvents(ctx, defers)
+	}()
+	return cart, err
+}
+
+// UpdateAdditionalDataForDelivery of cart
+func (cs *CartService) UpdateAdditionalDataForDelivery(ctx context.Context, session *web.Session, deliveryCode string, key string, value string) (*cartDomain.Cart, error) {
```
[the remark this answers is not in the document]
carstendietrich: also, I would expect to see a map[string]string here as param.

cart/interfaces/graphql/dto/additional_data.go
```diff
@@ -0,0 +1,72 @@
...
+// Get attribute by key
+func (c *CustomAttributes) Get(key string) CustomAttribute {
+	if c.Attributes == nil {
+		return nil
+	}
+
+	if value, found := c.Attributes[key]; found {
+		var attribute CustomAttribute
+
+		switch value {
```
carstendietrich, line 38: why not using: `strconv.ParseBool(value)` here?

cart/interfaces/graphql/mutationresolvers.go
```diff
@@ -252,6 +252,34 @@ func (r *CommerceCartMutationResolver) CartClean(ctx context.Context) (bool, err
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return r.q.CommerceCart(ctx)
+}
+
+// UpdateAdditionalDataForDeliveries of delivery info
+func (r *CommerceCartMutationResolver) UpdateAdditionalDataForDeliveries(ctx context.Context, additionalDataList []*dto.DeliveryAdditionalData) (*dto.DecoratedCart, error) {
```
carstendietrich, line 269: Would go here for `UpdateDeliveriesAdditionalData`

cart/interfaces/graphql/mutationresolvers.go
```diff
@@ -252,6 +252,34 @@ func (r *CommerceCartMutationResolver) CartClean(ctx context.Context) (bool, err
 	return true, nil
 }
 
+// UpdateAdditionalData of cart
+func (r *CommerceCartMutationResolver) UpdateAdditionalData(ctx context.Context, additionalDataList []*dto.KeyValue) (*dto.DecoratedCart, error) {
+	session := web.SessionFromContext(ctx)
+	for _, additionalData := range additionalDataList {
+		_, err := r.cartService.UpdateAdditionalData(ctx, session, additionalData.Key, additionalData.Value)
```
carstendietrich, line 259: This will lead to a lot of backend calls, e.g. 10 key:value pairs will lead to 10 calls. See `CartService` comments to have a function to update all additional data at once.

cart/interfaces/graphql/schema.graphql
```diff
@@ -463,4 +487,8 @@ extend type Mutation {
     Commerce_Cart_UpdateDeliveryShippingOptions(shippingOptions: [Commerce_Cart_DeliveryShippingOption!]): [Commerce_Cart_DeliveryAddressForm]!
     "Cleans current cart"
     Commerce_Cart_Clean: Boolean!
+    "Adds/Updates additional data for the cart"
+    Commerce_Cart_UpdateAdditionalData(additionalData: [Commerce_Cart_KeyValue!]!): Commerce_DecoratedCart!
+    "Adds/Updates additional data for the given deliveries"
+    Commerce_Cart_UpdateAdditionalDataForDeliveries(data: [Commerce_Cart_DeliveryAdditionalData!]!): Commerce_DecoratedCart!
```
carstendietrich, line 493: here then `Commerce_Cart_UpdateDeliveriesAdditionalData`

carstendietrich (changes requested):

cart/interfaces/graphql/dto/additional_data.go
el0c1n: because it should not convert the strings 0 or 1 as booleans in case these values are saved as non boolean attributes

carstendietrich: Mhh okay, in general, I kinda dislike the approach when you put in a string and get a bool back. Especially because the mutation only accepts strings but then can return bool. Do we really need that type casting or should that be done in the consumer?

PR #318, opened 2021-05-26
GraphQL: adjust response type of update delivery shipping options mutation

lifedraft:
`Commerce_Cart_UpdateDeliveryShippingOptions` mutation responded with slice of `Commerce_Cart_DeliveryAddressForm` which was incorrect as we don't process any form data within the mutation. It responds now rightly only with `processed` state.


cart/interfaces/graphql/mutationresolvers.go
```diff
@@ -199,32 +199,17 @@ func (r *CommerceCartMutationResolver) CommerceCartUpdateDeliveryAddresses(ctx c
...
 		delivery, found := cart.GetDeliveryByCode(shippingOption.DeliveryCode)
 		if !found {
-			formResult.ValidationInfo.GeneralErrors = []domain.Error{
-				{
-					MessageKey: "delivery_not_found",
-				},
-			}
-			result = append(result, &formResult)
-			continue
+			return nil, errors.New("delivery not found")
```
tessig, line 212:
```suggestion
			return nil, cartDomain.ErrDeliveryCodeNotFound
```

cart/interfaces/graphql/mutationresolvers.go
```diff
@@ -235,11 +220,9 @@ func (r *CommerceCartMutationResolver) CommerceCartUpdateDeliveryShippingOptions
 		if err != nil {
 			return nil, err
 		}
-		formResult.Processed = true
-		result = append(result, &formResult)
 	}
 
-	return result, nil
+	return &dto.UpdateShippingOptionsResult{Processed: true}, nil
```
tessig, line 225: Do we need the `Processed` field if it's always true?

tessig (changes requested):

cart/interfaces/graphql/mutationresolvers.go
lifedraft: Thought about just resolving with true and decided to be keep the return type with one and only process field so we can extend it in the future.

tessig (approved):

tessig (approved):

PR #330, opened 2021-08-10
checkout: reserve a new order id if early place order has payment issues

carstendietrich:
If the payment provider enforces an early place order and the payment is then aborted we never renew the order id.


tessig (approved):
looks ok.
we should perhaps rethink the whole order controller ;)

PR #337, opened 2021-09-16
payment: introduce client-sdk payment method

tessig:
client sdk payment is meant to be implemented completely in the frontend


carstendietrich (approved):
LGTM

PR #338, opened 2021-09-16
Change license to MIT

tessig:
As discussed with the team, future releases of flamingo-commerce will be under the MIT License like flamingo-core


carstendietrich (approved):
LGTM

PR #351, opened 2021-11-30
cart: dispatch events before and after cart merge for e.g. logging

ChristopherSchwab:


cart/application/eventHandling.go
```diff
@@ -20,6 +20,18 @@ type (
 		cartReceiverService *CartReceiverService
 		cartCache           CartCache
 		webIdentityService  *auth.WebIdentityService
+		eventRouter         flamingo.EventRouter
+	}
+
+	// PreCartMergeEvent is dispatched after getting the (current) guest cart and the customer cart before merging
+	PreCartMergeEvent struct {
+		GuestCart    *cartDomain.Cart
```
carstendietrich, line 28: Are the carts passed by reference for a specific reason? would change that to value objects to avoid accidentally mutating the structs in the event listener.

carstendietrich (approved):
LGTM

PR #352, opened 2022-01-11
product: add separate fake for livesearch

tessig:


product/module.go
```diff
@@ -77,7 +78,8 @@ commerce: {
 			enabled: bool | *false
 			currency: *"€" | !=""
 			if enabled {
-			  jsonTestDataFolder?: string | !=""
+			  jsonTestDataFolder: string | !=""
```
carstendietrich, line 81: Are those fields now really mandatory? Thought that the live search is optional as well as the test data folder in general..

carstendietrich (changes requested):
Besides the bc breaking config changes fine for me.

carstendietrich (approved):

PR #356, opened 2022-01-25
product: Add category facet to the fake product search service

el0c1n:


product/infrastructure/fake/helper.go
```diff
@@ -60,3 +64,36 @@ func unmarshalJSONProduct(productRaw []byte) (domain.BasicProduct, error) {
 
 	return *simpleProduct, nil
 }
+
+//go:embed testdata
```
tessig, line 68: seems, that we just need `testdata/categoryFacetItems.json`, so we could embed that file directly

tessig (approved):

PR #401, opened 2022-09-14
Draft: feat(product): add extended stock attribute

KarolNowakk:


product/domain/productBasics.go
```diff
@@ -63,7 +63,9 @@ type (
 
 		CategoryToCodeMapping []string
 
+		// Deprecated: use Stock[x].Level instead
 		StockLevel string
```
tessig, line 67: We should also refactor all usages of this field. see product/infrastructure/fake/productservice.go

product/domain/productBasics.go
```diff
@@ -570,6 +580,7 @@ func findMediaInProduct(p BasicProduct, group string, usage string) *Media {
 
 // IsInStock returns information if current product whether in stock or not
 func (bpd BasicProductData) IsInStock() bool {
+	// todo: not sure what to do here
```
tessig, line 583: imho we should check the new field first and only if not set fall back to the deprecated field.
The logic would be: if any element of the stock array has InStock: true

In Addition we can add `IsInStockForDeliveryCode(deliveryCode string) bool`

tessig (changes requested):

product/domain/productBasics.go
```diff
@@ -570,13 +580,30 @@ func findMediaInProduct(p BasicProduct, group string, usage string) *Media {
...
 
+// IsInStockForDeliveryCode returns information if current product whether in stock or not for provided delivery code
+func (bpd BasicProductData) IsInStockForDeliveryCode(deliveryCode string) bool {
+	for _, stock := range bpd.Stock {
+		if stock.DeliveryCode == deliveryCode {
+			return stock.InStock
+		}
+	}
+
+	return true
```
tessig, line 604: I would say out of stock when we do not have info for the deliveryCode
```suggestion
	return false
```

tessig (changes requested):

tessig (approved):

PR #404, opened 2022-10-10
Draft: feat(sourcing): add source error

KarolNowakk:


carstendietrich:
lets do this inside our concrete implementation

PR #405, opened 2022-10-10
feat(sourcing): enhance error message when no source available

KarolNowakk:


sourcing/domain/service.go
```diff
@@ -115,6 +116,7 @@ func (d *DefaultSourcingService) GetAvailableSources(ctx context.Context, produc
 
 	sources, err := d.availableSourcesProvider.GetPossibleSources(ctx, product, deliveryInfo)
 	if err != nil {
+		d.logger.Warn("no possible sources available: " + err.Error())
```
carstendietrich, line 119: let's not log here, if the outer layer wants to log they can use the error that we bubble up to do so..

carstendietrich (changes requested):
lgtm, only minor logging thingy

PR #414, opened 2022-11-10
feat(checkout/placeorder): Make canceling orders in rollback configurable

carstendietrich:
In case of an error, for example, during the payment process, it is often not desirable to cancel the order but to leave it in the state with open payment. Otherwise, it could appear that the user has deliberately canceled the order. 

This MR changes the default behavior and makes it optionally possible to cancel the order by configuration.

### DIFF
--- a/checkout/module.go
+++ b/checkout/module.go
@@ -106,6 +106,7 @@ func (m *Module) Configure(injector *dingo.Injector) {
 
 // CueConfig definition
 func (m *Module) CueConfig() string {
+	// language=cue
 	return `
 commerce: checkout: {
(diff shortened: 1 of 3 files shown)


checkout/module.go
```diff
@@ -138,6 +139,11 @@ commerce: checkout: {
 				redis: Redis
 			}
 		}
+		states: {
+			placeorder: {
+				cancelOrdersDuringRollback: bool | *false
```
tessig, line 144: With default value `true` the change would not be breaking

tessig (approved):

PR #432, opened 2023-02-07
chore: bump to go.19 and mockery compatibility for go 1.20

tessig:

PR #440, opened 2023-03-06
chore: downgrade redigo version to v1.8.9, upgrade go version to 1.19

IvanMaidurov:


carstendietrich:
maybe have a look at #432 guess @tessig started the go update already there.

PR #441, opened 2023-03-06
feat(cart): add support for AdditionalData on CartItem level for the defaultCartBehaviour

lifedraft:
Currently `updateItem` on the `defaultCartBehavior` (Memory Cart) is lacking support for setting the AdditionalData. This PR is adding AdditionalData from the Update command to the item.


tessig:
The problem with go generate should be fixed soon in master. Please wait for the fix and rebase afterwards. Thanks

PR #449, opened 2023-03-28
feat(cart): update default cart behaviour to support other cart storage implementations

ChristopherSchwab:


cart/infrastructure/defaultCartBehaviour.go
```diff
@@ -77,14 +97,16 @@ func (cob *DefaultCartBehaviour) Inject(
 		DefaultCurrency string  `inject:"config:commerce.cart.defaultCartAdapter.defaultCurrency"`
 	},
 ) {
-	cob.cartStorage = CartStorage
-	cob.productService = ProductService
-	cob.logger = Logger.WithField(flamingo.LogKeyCategory, "inmemorybehaviour")
+	cob.cartStorage = cartStorage
+	cob.productService = productService
+	cob.logger = logger.WithField(flamingo.LogKeyCategory, "inmemorybehaviour")
```
carstendietrich, line 102: that seems to be off, and please move the `.WithFields` to each logger call because we otherwise waste memory.

cart/infrastructure/defaultCartBehaviour.go
```diff
@@ -97,96 +119,80 @@ func (cob *DefaultCartBehaviour) Complete(ctx context.Context, cart *domaincart.
...
+
 	return cart, nil, nil
 }
 
 // Restore supplied cart (implements CompleteBehaviour)
 func (cob *DefaultCartBehaviour) Restore(ctx context.Context, cart *domaincart.Cart) (*domaincart.Cart, domaincart.DeferEvents, error) {
 	newCart, err := cart.Clone()
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, errors.Wrap(err, "cart.infrastructure.DefaultCartBehaviour: error on cloning cart")
```
carstendietrich, line 130: please use `fmt.Errorf("there is a issue: %w", err)`, and drop the `"github.com/pkg/errors"` import since that package is deprecated.

cart/infrastructure/defaultCartBehaviour.go
```diff
@@ -54,21 +57,38 @@ type (
...
 	// DefaultVoucherHandler implements a basic voucher handler
 	DefaultVoucherHandler struct{}
+
+	CartNotFoundError struct {
+		action string
+		id     string
+	}
 )
 
+func (e *CartNotFoundError) Error() string {
```
carstendietrich, line 67: would prefer to have a regular error like:
```
var ErrCartNotFound = errors.New("cart not found")
```

and wrapping it then in places like `fmt.Errorf("%w for cart id %q, during action", err, cart.ID, action)`

cart/infrastructure/defaultCartBehaviour.go
```diff
@@ -827,61 +892,22 @@ func (cob *DefaultCartBehaviour) collectTotals(cart *domaincart.Cart) {
 	}
 }
 
-// ApplyVoucher checks the voucher and adds the voucher to the supplied cart if valid
-func (DefaultVoucherHandler) ApplyVoucher(_ context.Context, cart *domaincart.Cart, couponCode string) (*domaincart.Cart, error) {
```
carstendietrich, line 831: should we move this to some sort of FakeVoucherHandler guess some of the integration tests are relying on this behaviour, same for giftcards

cart/infrastructure/defaultCartBehaviour.go
```diff
@@ -54,21 +57,38 @@ type (
...
 
 // Inject dependencies
 func (cob *DefaultCartBehaviour) Inject(
-	CartStorage CartStorage,
-	ProductService domain.ProductService,
-	Logger flamingo.Logger,
+	cartStorage CartStorage,
+	productService domain.ProductService,
+	logger flamingo.Logger,
 	voucherHandler VoucherHandler,
```
carstendietrich, line 92: not sure if to mark it as optional or not.

PR #462, opened 2023-05-17
fix(sourcing): fix finding source with least available qty

KarolNowakk:


sourcing/domain/service.go
```diff
@@ -13,6 +13,8 @@ import (
 	"flamingo.me/flamingo/v3/framework/flamingo"
 )
 
+const MaxSourceQyt = math.MaxInt64
```
carstendietrich, line 16: typo

sourcing/domain/service.go
```diff
@@ -452,9 +454,11 @@ func (d *DefaultSourcingService) getSourceStock(
 func (s AvailableSources) QtySum() int {
 	qty := 0
 	for _, sqty := range s {
-		if sqty == math.MaxInt64 {
-			return sqty
+		// check against max int 32 to avoid overflowing int
+		if sqty == MaxSourceQyt || qty > math.MaxInt32 {
```
carstendietrich, line 458: can you explain the second check for max int 32?

carstendietrich (changes requested):

sourcing/domain/service.go
KarolNowakk: package main

import (
	"fmt"
	"math"
)

func main() {
	sourceqty := int64(math.MaxInt64) - int64(1)
	sourceqty1 := int64(math.MaxInt64) - int64(1)

	sum := sourceqty + sourceqty1

	fmt.Println(sum) // -4
}

if fore some reason sqource qty will be almost maxint64 and not exactly and we sum it up we will have a wierd number like -4 in this example

PR #465, opened 2023-05-31
fix(cart): add item id to context before restrictor check

IvanMaidurov:


cart/application/cartService.go
```diff
@@ -62,8 +62,24 @@ type (
...
+	itemIDKey contextKeyType = "item_id"
 )
 
+func ItemIDFromContext(ctx context.Context) string {
+	id, _ := ctx.Value(itemIDKey).(string)
+
+	return id
+}
+
+func contextWithItemID(ctx context.Context, itemID string) context.Context {
```
carstendietrich, line 79: So we don't plan on making this public? So when i want to call the restrictor from places else then cart service?

PR #469, opened 2023-08-09
feat(cart): distinguish country codes in the phone number

IvanMaidurov:


docs/openapi/docs.go
```diff
@@ -1601,7 +1587,7 @@ var doc = `{
                     }
                 },
                 "BundleConfig": {
-                    "$ref": "#/definitions/cart.BundleConfiguration"
```
tessig, line 1604: This looks unexpected. Could you please double-check that this is wanted?

IvanMaidurov: this one correct

tessig (approved):

PR #334, opened 2021-08-23
Remove github.com/pkg/errors in favor of Go core errors package

carstendietrich:

PR #593, opened 2024-08-01
chore: Replace github.com/pkg/errors with Go core error package

carstendietrich:
The topic was started in 2021 (see #334) and is now actually in a mergeable state.. 🎉

### DIFF
--- a/cart/domain/cart/cart.go
+++ b/cart/domain/cart/cart.go
@@ -275,7 +275,7 @@ func (c Cart) GetDeliveryByItemID(itemID string) (*Delivery, error) {
 		}
 	}
 
-	return nil, fmt.Errorf("delivery not found for %q", itemID) //nolint:err113 //not worth introducing a dedicated error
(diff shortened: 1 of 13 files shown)


cart/domain/cart/cart.go
```diff
@@ -276,7 +275,7 @@ func (c Cart) GetDeliveryByItemID(itemID string) (*Delivery, error) {
 		}
 	}
 
-	return nil, errors.Errorf("delivery not found for %q", itemID)
+	return nil, fmt.Errorf("delivery not found for %q", itemID) //nolint:err113 //not worth introducing a dedicated error
```
IvanMaidurov, line 278: looks to my like not found with different comments, why not worth?

carstendietrich: valid point, updated the code to reuse existing errors instead of muting the linter..

IvanMaidurov (approved):

PR #722, opened 2026-05-18
chore: fix add to cart

KarolNowakk:


cart/infrastructure/defaultCartBehaviour.go
```diff
@@ -103,6 +106,15 @@ func (cob *DefaultCartBehaviour) Inject(
 	}
 }
 
+// matcher returns the configured CartItemMatcher, falling back to the
+// default implementation when no matcher was injected (e.g. in unit tests).
+func (cob *DefaultCartBehaviour) matcher() CartItemMatcher {
```
tessig: Do we need this? It is bound in the module.go anyway, and imho it is not worth this extra logic "just for" tests. Dingo will assure that it is injected in production code.

tessig (approved):

PR #723, opened 2026-06-16
feat(prices): add to and from times to the price context

IvanMaidurov:


product/domain/productBasics.go
```diff
@@ -144,6 +144,8 @@ type (
 		CustomerGroup string
 		ChannelCode   string
 		Locale        string
+		From          string
+		To            string
```
tessig: Shouldn't that be time?

IvanMaidurov: we use both for time fields, but change to time.

tessig (approved):

PR #724, opened 2026-06-22
Add cancellation reason support for checkout and payment workflows

kgrigorev:
### Description

This pull request introduces an optional payment **cancellation reason** that is threaded end-to-end through the checkout and payment workflows. The reason is optional at every layer and defaults to `Unspecified`, preserving existing (legacy) behavior and keeping the change backward-compatible.

#### Checkout
- Added a new GraphQL enum `Commerce_Checkout_PaymentCancellationReason` (`UNSPECIFIED`, `ABORTED_BY_CUSTOMER`) and an optional `reason` argument to the `Commerce_Checkout_CancelPlaceOrder` mutation.
- Added a `dto.PaymentCancellationReason` type plus `MapPaymentCancellationReason` (nil/unknown → `Unspecified`), wired through the mutation resolver and registered via `types.Map`.
- Threaded the reason from the resolver into `CancelPlaceOrderCommand` → `Handler.CancelPlaceOrder` → `Coordinator.Cancel`.
- Carried the reason through the process context (new `Context.CancelReason`) and the rollback mechanism, so state rollbacks receive it via the context hop.
- Failure reasons (`CanceledByCustomerReason`, `PaymentCanceledByCustomerReason`) now carry the cancellation reason; the payment-aborted validation path now reports `ABORTED_BY_CUSTOMER` instead of an empty reason.

#### Payment
- Introduced `CancellationReason` as a domain enum for a consistent representation (zero value `Unspecified` = legacy behavior).
- Added the `reason` parameter to the `WebCartPaymentGateway.CancelOrderPayment` interface and updated all implementations and mocks (`OfflineWebCartPaymentGateway`, the projecttest fake gateway, and the generated mock).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@
 
 **checkout**
 * Added support for login, password and tls to the redis client
+* Enhanced the cancellation flow with detailed reasons threading through the place-order process
+* GraphQL: Added an optional `reason` argument to the `Commerce_Checkout_CancelPlaceOrder` mutation
+
+**payment**
+* **Breaking:** `WebCartPaymentGateway.CancelOrderPayment` now requires a `CancellationReason` argument. Added the `CancellationReason` type to the domain.
 
 **search**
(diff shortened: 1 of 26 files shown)


tessig (approved):

tessig (dismissed):
Please add a changelog entry

tessig (approved):

PR #729, opened 2026-06-29
fix(checkout): wrap payment validation error in custom error

dylanprovost:
The current payment validation risks exposing directly errors from the payment provider to the client. Here is a proposal to return custom flamingo-commerce errors instead.

### DIFF
--- a/checkout/application/placeorder/validate_payment.go
+++ b/checkout/application/placeorder/validate_payment.go
@@ -19,6 +19,10 @@ const (
 	ValidatePaymentErrorNoWalletDetails = "no wallet details set for action"
 	// ValidatePaymentErrorNoActionDisplayData used for errors when the needed DisplayData/HTML is missing from the ActionData struct
 	ValidatePaymentErrorNoActionDisplayData = "no display data / html set for action"
+	// ValidatePaymentErrorStatusFailed used for errors when the payment status is failed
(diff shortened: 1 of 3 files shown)


checkout/application/placeorder/validate_payment.go
```diff
@@ -103,11 +107,21 @@ func PaymentValidator(ctx context.Context, p *process.Process, paymentService *a
...
+			err = ValidatePaymentErrorStatusFailed
 		}
+
+		return process.RunResult{
+			Failed: process.PaymentErrorOccurredReason{Error: err},
+		}
+	case paymentDomain.PaymentFlowStatusCancelled:
+		err := ""
+		if flowStatus.Error != nil {
+			err = ValidatePaymentErrorStatusCancelled
```
IvanMaidurov: it seems like error check is not needed here? If there is no error we fail with empty error message?

dylanprovost: Yes, I did not want to break the previous logic, which was doing exactly that. I do not know if there is any case where "failed" or "cancelled" should **not** return any error. 
Happy to change it if it did not make sense in the first place.

IvanMaidurov: honestly like it way more after your changes, what do you think @tessig ?

tessig (approved):

